PR #40, opened 2020-04-13
Update version and layers.json via script

tianchu:
### What does this PR do?

Use `yarn version` to bump the version number and commit it with the layers.json changes as part of the release script. 

### Testing Guidelines

How did you test this pull request?

### Additional Notes

Anything else we should know when reviewing?

### DIFF
--- a/scripts/publish_prod.sh
+++ b/scripts/publish_prod.sh
@@ -28,17 +28,40 @@ if [ -z "$AWS_SESSION_TOKEN" ]; then
     exit 1
 fi
 
+# Read the current version
+CURRENT_VERSION=$(node -pe "require('./package.json').version")
+
+# Read the desired version
+if [ -z "$1" ]; then
+    echo "Must specify a desired version number"
+    exit 1
+elif [[ ! $1 =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+    echo "Must use a semantic version, e.g., 3.1.4"
+    exit 1
+else
+    VERSION=$1
+fi
+
+# Confirm to proceed
+read -p "About to bump the version from ${CURRENT_VERSION} to ${VERSION}, and publish. Continue (y/n)?" CONT
+if [ "$CONT" != "y" ]; then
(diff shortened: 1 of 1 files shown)


scripts/publish_prod.sh
```diff
@@ -28,17 +28,39 @@ if [ -z "$AWS_SESSION_TOKEN" ]; then
...
+    exit 1
+elif [[ ! $1 =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+    echo "Must use a semantic version, e.g., 3.1.4"
+    exit 1
+else
+    VERSION=$1
+fi
+
+# Confirm to proceed
+read -p "About to bump the version from ${CURRENT_VERSION} to ${VERSION}, and publish. Continue (y/n)?" CONT
```
agocs, line 46: Is it worth thinking about how to add a `-f` flag so that this can be used by SDP for automatic releases, or should we worry about that in Q3?

tianchu: I think we worry about that later, we may need to replace this script with something completely different to configure the release process.

agocs: 👍

scripts/publish_prod.sh
```diff
@@ -28,17 +28,40 @@ if [ -z "$AWS_SESSION_TOKEN" ]; then
...
 yarn build
-yarn publish --new-version "$PACKAGE_VERSION"
+yarn publish --new-version "$VERSION"
 
-echo 'Tagging Release'
-git tag "v$PACKAGE_VERSION"
-git push origin "refs/tags/v$PACKAGE_VERSION"
\ No newline at end of file
+echo 'Pushing updates to github'
+git push origin master
```
DarcyRaynerDD, line 66: We will need to disable branch protection on master for this to work

tianchu: I believe admins are allowed for direct (non-forced) pushes by default? I can try it out.

PR #45, opened 2020-05-07
Add support for setting forwarder arn directly in plugin

DarcyRaynerDD:
### What does this PR do?

* Now possible to specify the forwarder arn directly in the serverless.yml, making manual trigger setup easier.
* Also updates dependencies, (sorry for the big diff).

### Motivation

What inspired you to submit this pull request?

### Testing Guidelines

How did you test this pull request?

### Additional Notes

Anything else we should know when reviewing?


README.md
```diff
@@ -77,6 +77,10 @@ custom:
 
     # Enable tracing on Lambda function using dd-trace, datadog's APM library. Requires datadog log forwarder to be set up. Defaults to true.
     enableDDTracing: true
+
+    # When set, the plugin will subscribe the lambda's cloudwatch log groups to the forwarder with the
+    # given arn. Don't use this if you are already using Datadog's automatic trigger management.
```
tianchu, line 82: Does this conflict with the automatic trigger management somehow? Otherwise I would delete _"Don't use this if you are already using Datadog's automatic trigger management."_ In fact, I would treat the automatic trigger management as a "catch all" or "fallback" solution for ad-hoc functions, and for functions created by the serverless framework i'd prefer have the trigger defined in serverless.yml (even when automatic trigger management is configured).

src/forwarder.ts
```diff
@@ -0,0 +1,34 @@
...
+const logGroupSubscriptionKey = "AWS::Logs::SubscriptionFilter";
+
+interface LogGroupResource {
+  Type: typeof logGroupKey;
+  Properties: {
+    LogGroupName: string;
+  };
+}
+
+function isLogGroup(value: any): value is LogGroupResource {
```
tianchu, line 13: We probably should narrow the check to `isLambdaLogGroup`, since log groups could be created for other services through serverless.yml? I cannot think of an example right away, but I guess it doesn't hurt to be more strict anyway?

DarcyRaynerDD: I guess the step function log group could be created in this way, since you have to manually create the log group. I'll add a check on the log group name

README.md
DarcyRaynerDD: Yeah, it can conflict. Essentially, if you are using automatic trigger management, and it's already subscribed to your function's log group, then trying to set the subscription will fail, since you can't have two subscriptions. I actually ran into this issue during testing. For new functions it's fine though.

src/forwarder.ts
```diff
@@ -0,0 +1,34 @@
...
+  return value.Type === logGroupKey;
+}
+
+export function addCloudWatchForwarderSubscriptions(service: Service, functionArn: string) {
+  const resources = service.provider.compiledCloudFormationTemplate.Resources;
+
+  for (const [name, resource] of Object.entries(resources)) {
+    if (!isLogGroup(resource) || !resource.Properties.LogGroupName.startsWith("/aws/lambda/")) {
```
tianchu, line 21: nits, does it make the code more explicit by moving the `Properties.LogGroupName.startsWith("/aws/lambda/")` check into the helper function and rename the helper function to `isLambdaLogGroup`?

DarcyRaynerDD: The other function is a type guard, so it's just being used to confirm the structure of the object is correct. Since the '/aws/lambda' isn't really encoded into the interface/type, my thinking was it shouldn't live inside the type guard.

README.md
DarcyRaynerDD: I've reworded this section, and added a check before we try to apply the subscription to the arn, so it's less likely a conflict will occur.

PR #48, opened 2020-06-03
Automatically tagging service and env

hannah-sun:
### What does this PR do?

* By default, the plugin automatically tags the lambda functions with `service` and `env` using the service name and stage values.
* This will not override existing tags and can be turned off. 

### Motivation
To better organize serverless observability data using `service` and `env` tags.

### Testing Guidelines
Added some unit tests and tested on sandbox.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -31,8 +31,10 @@ export interface FunctionDefinitionWithTags {
   tags?: { [key: string]: string };
 }
 
-const serviceKey = "service";
-const envKey = "env";
+enum TagKeys {
+  Service = "service",
+  Env = "env",
+}
 
 module.exports = class ServerlessPlugin {
   public hooks = {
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -118,4 +140,24 @@ module.exports = class ServerlessPlugin {
       }
     }
   }
+
+  private handleTags() {
```
sfirrin, line 144: I'd probably name this function something like `addTags` or `addServiceAndEnvTags` just to make it a little more clear what it's doing at a glance

src/index.ts
```diff
@@ -16,6 +16,23 @@ import { cleanupHandlers, writeHandlers } from "./wrapper";
...
+  name: string;
+  package: Package;
+  runtime?: string;
+  handler: string;
+  timeout?: number;
+  memorySize?: number;
+  environment?: { [name: string]: string };
+  events: Event[];
+  tags?: { [key: string]: string };
+}
```
sfirrin, line 32: I think it would probably be preferable to build the fixed interface off the incomplete DefinitelyTyped interface with something like
```
interface FunctionDefinitionWithTags extends FunctionDefinition {
  tags: {[key: string]: string}
}
```
just to keep as much as possible in line with the official types

DarcyRaynerDD:
LGTM 👍

PR #50, opened 2020-06-11
Adopt code free instrumentation

hannah-sun:
### What does this PR do?
Updating the plugin to adopt a new approach for enabling instrumentation without customers changing their code.

### Motivation
The python and js lambda layers have been changed so customers can apply the datadog wrapper for instrumentation without changing their functions code. We want to update the plugin to adopt this approach as well.

### Testing Guidelines
Updated and added unit tests, ran test serverless deployment on sandbox.


src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
-
-export async function writeHandlers(service: Service, funcs: FunctionInfo[], tracingMode: TracingMode) {
-  await cleanupHandlers();
-  await util.promisify(fs.mkdir)(datadogDirectory);
-  const groups = getFunctionGroups(funcs);
-  const promises = groups.map(async (funcGroup) => {
-    const methods = getMethods(funcGroup);
-    const result = getWrapperText(funcGroup.runtime, funcGroup.filename, methods, tracingMode);
-    if (result === undefined) {
+export function redirectHandlers(service: Service, funcs: FunctionInfo[], addLayers: boolean) {
```
sfirrin, line 19: I think it'd be helpful to have a docstring above this function to briefly explain what it's doing

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
-import { getHandlerPath, removeDirectory } from "./util";
-import { TracingMode } from "./templates/common";
 
-export const datadogDirectory = "datadog_handlers";
+export const datadogHandlerEnvVar = "DD_LAMBDA_HANDLER";
+export const pythonHandler = "datadog_lambda.handler.handler";
+export const pythonHandlerFile = "datadog_lambda.handler.py";
+export const jsHandlerLayerPrefix = "/opt/nodejs/";
+export const jsHandler = "node_modules/datadog-lambda-js/handler.handler";
+export const jsHandlerFile = "node_modules/datadog-lambda-js/handler.js";
```
sfirrin, line 17: nit: since these are globals it would be good to name them in all caps snake case to match the convention

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
-export async function writeHandlers(service: Service, funcs: FunctionInfo[], tracingMode: TracingMode) {
-  await cleanupHandlers();
-  await util.promisify(fs.mkdir)(datadogDirectory);
-  const groups = getFunctionGroups(funcs);
-  const promises = groups.map(async (funcGroup) => {
-    const methods = getMethods(funcGroup);
-    const result = getWrapperText(funcGroup.runtime, funcGroup.filename, methods, tracingMode);
-    if (result === undefined) {
+export function redirectHandlers(service: Service, funcs: FunctionInfo[], addLayers: boolean) {
+  const handlerFiles: Set<string> = new Set();
```
sfirrin, line 20: nit: you can also do `const handlerFiles = new Set<string>()`

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
-  addToExclusionList(service, allFiles);
+  addToExclusionList(service, [...handlerFiles]);
 }
 
-export async function cleanupHandlers() {
-  await removeDirectory(datadogDirectory);
-}
-
-export function getWrapperText(type: RuntimeType, filename: string, methods: string[], tracingMode: TracingMode) {
+function getDDHandler(type: RuntimeType | undefined, addLayers: boolean) {
```
sfirrin, line 45: nit: at first glance a variable named `type` could be almost anything, maybe something like `lambdaRuntime` would be better?

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
+  if (type === undefined) {
+    return;
+  }
+  const finalJsHandler = addLayers ? `${jsHandlerLayerPrefix}${jsHandler}` : jsHandler;
+  const finalJsHandlerFile = addLayers ? `${jsHandlerLayerPrefix}${jsHandlerFile}` : jsHandlerFile;
```
sfirrin, line 50: nit: it might be cleaner to put these within the JS case block so that if the function is Python the definitions get skipped

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
-  await cleanupHandlers();
-  await util.promisify(fs.mkdir)(datadogDirectory);
-  const groups = getFunctionGroups(funcs);
-  const promises = groups.map(async (funcGroup) => {
-    const methods = getMethods(funcGroup);
-    const result = getWrapperText(funcGroup.runtime, funcGroup.filename, methods, tracingMode);
-    if (result === undefined) {
+export function redirectHandlers(service: Service, funcs: FunctionInfo[], addLayers: boolean) {
+  const handlerFiles: Set<string> = new Set();
+  funcs.map((func) => {
```
sfirrin, line 21: nit: since we don't need the array that's getting built and returned by map I'd refactor `funcs.map` to `funcs.forEach`

src/wrapper.spec.ts
```diff
@@ -308,156 +40,133 @@ describe("writeHandlers", () => {
...
-      [datadogDirectory]: {
-        "unused-file": "some-value",
-      },
-    });
     const service = {} as any;
-    await writeHandlers(service, [], TracingMode.XRAY);
-    expect(fs.existsSync(`${datadogDirectory}/unused-file`)).toBeFalsy();
-  });
-
-  it("ignores handlers with poorly formatted names", async () => {
```
hannah-sun, line 346: Wasn't sure if we still want to ignore certain handlers, or just redirect all by default?

src/wrapper.ts
hannah-sun, line 17: Not sure on the path here, when I install the package with `npm`, my local path is `node_modules/datadog-lambda-js/dist/hander.js`, but after the changes of moving `redirectHandlers` into `afterPackageFunction`, no error is generated for either version of the file path.

src/index.ts
```diff
@@ -64,21 +65,8 @@ module.exports = class ServerlessPlugin {
...
-      case "node":
-        defaultNodeRuntime = RuntimeType.NODE;
-        break;
-    }
 
-    const handlers = findHandlers(this.serverless.service, defaultRuntime, defaultNodeRuntime);
+    const handlers = this.getHandlers(config.nodeModuleType);
     if (config.addLayers) {
       this.serverless.cli.log("Adding Lambda Layers to functions");
       this.debugLogHandlers(handlers);
```
hannah-sun, line 72: I left this unchanged as of right now, but two lines below this one here has:
```
if (hasWebpackPlugin(this.serverless.service)) {
        forceExcludeDepsFromWebpack(this.serverless.service);
}
```
I think we may have originally talked about removing this logic, but wanted to double check.

src/wrapper.ts
```diff
@@ -6,109 +6,63 @@
...
+        exclude: [],
+        include: [],
+      };
+    }
+    if (func.handler.package.include === undefined) {
+      func.handler.package.include = [];
     }
-    return `${filename}`;
+    func.handler.package.include.push(handlerFile);
+    handlerFiles.add(handlerFile);
```
DarcyRaynerDD, line 40: handlerFile points to a file in the layer right? Not sure if it makes sense to add it to the include list, since by the time we run this code the function has already been packaged. I don't think this has any effect.

src/index.ts
```diff
@@ -154,4 +140,21 @@ module.exports = class ServerlessPlugin {
       }
     });
   }
+
+  private getHandlers(nodeModuleType: "es6" | "node" | "typescript" | undefined) {
+    const defaultRuntime = this.serverless.service.provider.runtime;
+    let defaultNodeRuntime: RuntimeType.NODE | RuntimeType.NODE_ES6 | RuntimeType.NODE_TS | undefined;
+    switch (nodeModuleType) {
+      case "es6":
```
DarcyRaynerDD, line 148: The es6 and TS types were added to help us use different handler templates  when using webpack/typescript plugins. They all map to a regular node runtime. Not sure if they are necessary anymore

Issue #52, opened 2020-06-30
Plugin does not respect provider.tags configuration


TylerHendrickson:
## Expected Behavior

When `provider.stackTags` are set in Serverless, they should be used to configure tags on individual Lambda function resources within the stack, unless a conflicting tag is set using `provider.tags` or `functions.functionName.tags`, per the README documentation for the `enableTags` setting (emphasis added):

> When set, the plugin will try to automatically tag lambdas with service and env, **but will not override existing tags**. Defaults to true.

Therefore, when I configure Serverless like this:
```yaml
service:
  name: my-service

plugins:
  - serverless-plugin-datadog

custom:
  datadog:
    addLayers: true
    enableTags: true

provider:
  name: aws
  stage: my-stage
  tags:  # Can also be set using stackTags
    env: production
    service: some-service

functions:
  MyFunction:
    handler: functions.handler
    runtime: python3.8
```

I expect that the `MyFunction` Lambda would be tagged with `env: production` and `service: some-service`.

## Actual Behavior

The `MyLambda` plugin is tagged with `env:my-stage` and `service:my-service` because `custom.datadog.enableTags` is set to `true` (the default value). The only way I can ensure that `MyFunction` is tagged with `env: production` and `service: some-service` is if `functions.MyFunction.tags.env` and `functions.MyFunction.tags.service` are set explicitly. For services with more that one function, this quickly becomes a nuisance, especially since the `provider.tags` setting is specifically provided by the Serverless framework for this purpose.

## Steps to Reproduce the Problem

  1. Configure `provider.tags.env` to a value other than that of `provider.stage` and configure `provider.tags.service` to a value other than that of `service.name`.
  1. Do not set `tags.env` or `tags.service` on at least one Lambda function
  1. Ensure that `datadog.enableTags` is not set to `false`
  1. Deploy and inspect the tags of the deployed function(s). You will see that `env:` is set to the value of `provider.stage` and `service:` is set to the value of `service.name`, instead of the values configured at `provider.tags.env` and `provider.tags.service`, respectively.

## Specifications

  - Serverless Framework version: 1.73.1
  - Datadog Serverless Plugin version: 1.0.0
  - Lambda function runtime (Python 3.7, Node 10, etc.): Reproduced using `python3.8` and `node10` runtimes
  
## Stacktrace

  N/A

hannah-sun:
[...]

PR #53, opened 2020-07-01
Check provider level tags before auto tagging service & env

hannah-sun:
### What does this PR do?
Before automatically tagging `service` and `env` with the `service` and `stage` values from the serverless.yml file, check if these tags already exist on the provider level (either under `tags` or `stackTags`). If so, we do not override those values.

### Motivation
Current implementation only checks for existing tags on the function level, and will override values set on the provider level. 
https://github.com/DataDog/serverless-plugin-datadog/issues/52

### Testing Guidelines
Added unit tests and tested with sample lambda function on sandbox.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -286,19 +286,23 @@ describe("ServerlessPlugin", () => {
       expect(functionWithTags).toHaveProperty("tags", { env: "dev", service: "test" });
     });
 
-    it("does not override tags set under provider", async () => {
+    it("does not override tags set on provider level", async () => {
       const function_ = functionMock({});
       const functionWithTags: ExtendedFunctionDefinition = function_;
       const serverless = {
         cli: { log: () => {} },
         getProvider: awsMock,
         service: {
+          getServiceName: () => "my-service",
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -119,22 +119,36 @@ module.exports = class ServerlessPlugin {
   }
 
   private addServiceAndEnvTags() {
```
sfirrin, line 121: suggestion: it would be helpful to have a docstring here to explain that it's checking for existing service and env tags and not overwriting them

src/index.spec.ts
```diff
@@ -285,5 +285,43 @@ describe("ServerlessPlugin", () => {
       await plugin.hooks["after:package:createDeploymentArtifacts"]();
       expect(functionWithTags).toHaveProperty("tags", { env: "dev", service: "test" });
     });
+
+    it("does not override tags set on provider level", async () => {
```
sfirrin, line 289: 👍 nice test

PR #69, opened 2020-09-15
Tag Lambda function with sls plugin version

Czechh:
### What does this PR do?

Adds a tag to the Lambda functions that use the sls plugin with `dd_sls_plugin: v${version}`.

### Motivation

This will give us data to drive product decisions.  

### Testing Guidelines

Unit tests and manually tested

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,6 @@ export interface ExtendedFunctionDefinition extends FunctionDefinition {
 enum TagKeys {
   Service = "service",
   Env = "env",
-  ForwarderVersion = "dd_forwarder_version",
   Plugin = "dd_sls_plugin",
 }
 
@@ -93,9 +92,6 @@ module.exports = class ServerlessPlugin {
       for (const error of errors) {
         this.serverless.cli.log(error);
       }
-
-      this.serverless.cli.log("Adding forwarder metadata tags to functions");
-      await this.addForwarderTags();
     }
 
     this.addPluginTag();
--- a/src/index.ts
+++ b/src/index.ts
@@ -94,11 +94,10 @@ module.exports = class ServerlessPlugin {
       }
     }
 
-    this.addPluginTag();
-
     if (config.enableTags) {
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -95,6 +97,7 @@ module.exports = class ServerlessPlugin {
     if (config.enableTags) {
       this.serverless.cli.log("Adding service and environment tags to functions");
       this.addServiceAndEnvTags();
+      this.addPluginTag();
```
tianchu, line 100: I think we should add the plugin version tag regardless of `config.enableTags`

src/index.ts
```diff
@@ -154,4 +157,20 @@ module.exports = class ServerlessPlugin {
...
+  private async addPluginTag() {
+    this.serverless.cli.log(`Adding Plugin Version ${version}`);
+
+    this.serverless.service.getAllFunctions().forEach((functionName) => {
+      const functionDefintion: ExtendedFunctionDefinition = this.serverless.service.getFunction(functionName);
+      if (!functionDefintion.tags) {
+        functionDefintion.tags = {};
+      }
+
+      functionDefintion.tags[TagKeys.Plugin] = `v${version}`;
```
tianchu, line 173: don't think it's a big deal, but I think the entire tag string is not allowed to start with digit but the tag value i allowed to start with a digit (another common example is status_code:200) https://a.cl.ly/KourX2vN

Czechh: Interesting. While testing with just the semver, I noticed we were stripping out the **first digit**.

 

So this might be a bug on our end. @DarcyRaynerDD mentioned that we could fix the parsing function, but then it wouldn't be consistent with the lambda tags crawler. So I decided to avoid that rabbit hole by adding a `v` prefix

src/index.ts
Czechh: Good call.

tianchu (approved):

PR #77, opened 2020-10-27
Add CloudFormation Outputs for Datadog links and log into terminal

JacksonDavenport:
### What does this PR do?
Add CloudFormation Outputs for each Function in the template, with a value of the URL linking to the corresponding function in Datadog.

### Motivation
Help navigation into the Datadog platform relative to these functions

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply


src/output.ts
```diff
@@ -0,0 +1,59 @@
...
+    const key = `${outputPrefix}${functionKey}`.replace(/[^a-z0-9]/gi, "");
+    outputs[key] = {
+      Description: `See ${functionKey} in Datadog`,
+      Value: `https://app.${site}/functions/${functionName}:${region}:${awsAccount}:aws?source=sls-plugin`,
+    };
+  });
+}
+
+export async function printOutputs(serverless: Serverless) {
+  const stackName = `${serverless.service.getServiceName()}-${serverless.getProvider("aws").getStage()}`;
```
tianchu, line 31: Does serverless gives you a builtin method to get the CFN stackName, asking because I wonder whether the stack name **always** follow this pattern.

src/output.ts
```diff
@@ -0,0 +1,59 @@
...
+  const describeStackOutput = await serverless
+    .getProvider("aws")
+    .request(
+      "CloudFormation",
+      "describeStacks",
```
tianchu, line 36: It's might be rare, but do you need to deal with the case where this API call fails (e.g., lack of permission) gracefully? e.g., logging a warning message (or even fail silently) instead of causing an exception to be displayed?

tianchu (approved):
Looks pretty good to me, except for two minor comments.

src/output.ts
JacksonDavenport: There is one that I missed, that I'll get this updated to use that. That will also take into account if you manually specify the `provider.stackName`. 

https://github.com/serverless/serverless/blob/cbc5e3c4e199a08650b02def8d1f9e91a13e7844/lib/plugins/aws/lib/naming.js#L47-L54

src/output.ts
JacksonDavenport: I'm not too sure when it would fail if everything else was successful and it got to this `after:deploy:deploy` stage, however, just to be safe I added exception handling in the case it does. If it comes up it will fail silently and just skip the whole Datadog Monitoring section.

tianchu (approved):

PR #86, opened 2020-12-05
Add config to exclude function from Datadog

nholbrook:
### What does this PR do?

This PR adds a config option that allows users to specify functions to be excluded from Datadog.

### Motivation

Currently, there is no way to prevent Datadog from updating certain functions. One example of a scenario where this is a problem is when using this plugin with `serverless-plugin-warmup`. Datadog will update the handler of the warmer function created by this plugin, which prevents it from working.

### Testing Guidelines

I added an additional test case and updated all associated test cases.

### Additional Notes

This issue is documented [here](https://github.com/FidelLimited/serverless-plugin-warmup/issues/231).

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [X] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [X] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [X] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [X] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog

PR #87, opened 2020-12-07
Add config to exclude function from Datadog

DarcyRaynerDD:
See #86

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ To further configure your plugin, use the following custom parameters in your `s
 | `forwarder`          | Setting this parameter subscribes the Lambda functions' CloudWatch log groups to the given Datadog forwarder Lambda function. Required when `enableDDTracing` is set to `true`.                                                                                                                                                                                                                 |
 | `enableTags`         | When set, automatically tag the Lambda functions with the `service` and `env` tags using the `service` and `stage` values from the serverless application definition. It does NOT override if a `service` or `env` tag already exists. Defaults to `true`.                                                                                                                                      |
 | `injectLogContext`         | When set, the lambda layer will automatically patch console.log with Datadog's tracing ids. Defaults to `true`.                                                                                                                                      |
+| `exclude`         | When set, this plugin ignore all specified functions. Use this parameter if you have any functions that should not include Datadog functionality. Defaults to `[]`.                                                                                                                                      |
 
 To use any of these parameters, add a `custom` > `datadog` section to your `serverless.yml` similar to this example:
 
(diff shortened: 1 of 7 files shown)


Czechh (approved):

README.md
```diff
@@ -35,6 +35,7 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `enableTags`         | When set, automatically tag the Lambda functions with the `service` and `env` tags using the `service` and `stage` values from the serverless application definition. It does NOT override if a `service` or `env` tag already exists. Defaults to `true`.                                                                                                                                      |
 | `injectLogContext`         | When set, the lambda layer will automatically patch console.log with Datadog's tracing ids. Defaults to `true`.                                                                                                                                      |
+| `exclude`         | When set, this plugin ignore all specified functions. Use this parameter if you have any functions that should not include Datadog functionality. Defaults to `[]`.                                                                                                                                      |
```
nhinsch, line 38: `this plugin ignore all` -> `this plugin will ignore all`

Also, I think it's worth specifying that this should be a comma-separated list of function names

Finally, is it normal that we expect the parameter to include [] brackets? Would it make more sense to just ask for a comma-separated string?

nhinsch (approved):
Looks good, one suggestion

README.md
DarcyRaynerDD: I think it's easier to use a list/array, since that's supported natively by yaml, and we don't need to do any extra parsing ourselves. For instance:

```
exclude:
 - first-excluded-func
 - seconded-excluded-fun
```

and 

```
exclude: [first-excluded-func, second-excluded-func]
```

Are equivalent.

nhinsch: Got it, but could we interpolate the string into the brackets? Rather than asking the user to supply the brackets?

If not, we should make it clear in the description that the parameter value must include the brackets.

nhinsch: Oh, is this going to be something that the user types into the YML file themselves? In that case never mind. I think it's still worth mentioning it's a comma-separated list of function names.

DarcyRaynerDD: I updated the eample to use the regular yaml list syntax, rather than the bracket syntax.

andyharley:
Is this working for anyone else? Trying this out with version 2.10.0 but just using this version of the plugin results in:

`Value of property Variables must be an object with String (or simple type) properties`

The generated cloudformation for the template contains:

```
               "Environment": {
                    "Variables": {
                        "DD_SITE": "datadoghq.com",
                        "DD_LOG_LEVEL": "info",
                        "DD_FLUSH_TO_LOG": true,
                        "DD_TRACE_ENABLED": true,
                        "DD_LOGS_INJECTION": true,
                        "DD_EXCLUDED_FUNCTIONS": [],
                        "MAX_DB_CONNECTION_LIMIT": "1",
                        "DD_LAMBDA_HANDLER": "App/myhandler"
                    }
                },
```

lists/arrays don't appear to be supported by cloudformation here i think: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-environment.html

tianchu:
@andyharley This is indeed an issue, I'm going to revert this change and release a new version. Thanks for the report!

PR #90, opened 2020-12-17
fix: store DD_EXLUDED_FUNCTIONS as comma-separated strings

### LINKED ISSUE #88: `exclude´ still injects a function handler property, does not work with the new `image` property
Hi there.

I am developing a service where one of our lambdas will be running using a custom ECR image as per [AWS's announcement on container image support](https://aws.amazon.com/blogs/aws/new-for-aws-lambda-container-image-support/). The serverless framework supports this in the new 2.15 version and you can now specify `image` instead of `handler` per function. If you supply both, it fails, which is expected and wanted behavior.

However, the `serverless-plugin-datadog` injects a custom handler which makes serverless to fail the validation check before deploying.

I tried to include my function in `custom.datadog.exclude` but it seems to still inject the handler. I had to make a `foo` runtime which this plugin does not understand to mitigate the problem.

## Expected Behavior

a) exclude does not inject a handler to the function
b) this plugin does not try to inject handlers to functions specified with `image` and optionally gives a warning

## Actual Behavior

Serverless deployment fails with `Either "handler" or "image" property (not both) needs to be set on function "convert-docker".`

## Steps to Reproduce the Problem

  1. add `serverless-plugin-datadog` to your project
  2. add a new function `foo` with an `image` attribute
  3. add `foo` to `custom.datadog.exclude`

## Specifications

  - Serverless Framework version: 2.15.0
  - Datadog Serverless Plugin version: 2.9.0
  - Lambda function runtime (Python 3.7, Node 10, etc.): n/a
  
Let me know what you think! Future plans and discussions for helping users with datadog integrations in docker containers in AWS lambda would also be really interesting.

hawkaa:
AWS Cloudformation does not allow for environment variables to be stored as lists. This commit
stores the DD_EXCLUDED_FUNCTIONS environment variable as a string separated with ";" instead
of a list.

Fixes #88

### DIFF
--- a/src/env.spec.ts
+++ b/src/env.spec.ts
@@ -128,7 +128,7 @@ describe("setEnvConfiguration", () => {
         enableDDTracing: true,
         enableTags: true,
         injectLogContext: false,
-        exclude: ["dd-excluded-function"],
+        exclude: ["dd-excluded-function", "dd-excluded-function-2"],
       },
       service,
     );
@@ -142,7 +142,7 @@ describe("setEnvConfiguration", () => {
           DD_SITE: "datadoghq.eu",
           DD_TRACE_ENABLED: true,
           DD_LOGS_INJECTION: false,
-          DD_EXCLUDED_FUNCTIONS: ["dd-excluded-function"],
+          DD_EXCLUDED_FUNCTIONS: "dd-excluded-function;dd-excluded-function-2",
         },
       },
     });
@@ -159,7 +159,7 @@ describe("setEnvConfiguration", () => {
           DD_SITE: "datadoghq.eu",
           DD_TRACE_ENABLED: false,
           DD_LOGS_INJECTION: false,
-          DD_EXCLUDED_FUNCTIONS: ["dd-excluded-function"],
+          DD_EXCLUDED_FUNCTIONS: "dd-excluded-function",
         },
       },
     } as any;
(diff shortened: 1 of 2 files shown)


hawkaa:
Ah. Just saw that @tianchu reverted the exclude functionality in #89 ...

tianchu:
@hawkaa Thanks for your PR! I think the better fix is to actually remove those lines to set the `DD_EXCLUDED_FUNCTIONS` environment variable, because the value of `exclude` is only needed by the plugin itself. Lmk if you want to give it a shot or wait for our fix (may take some time given the upcoming holidays). Thanks!

hawkaa:
I actually thought about that myself and I really didn't know why that information was stored in the environment in the first place. But I decided to go the "safe" option in case it was needed somewhere, although I found no other references.

+1 on removing the environment variable altogether. I might have time now before Christmas to have a look. For now, adding a custom runtime `foo` solves the problem for my dockerized lambda.

I am also really curious to see how we can facilitate the datadog plugin with the docker container lambdas. We lost our enhanced lambda metrics with our change, but I guess it should be easy for us to add it back as a python package in the docker image.

tianchu:
@hawkaa We did add [instructions](https://docs.datadoghq.com/serverless/installation/python?tab=containerimage) for Lambda functions deployed as container images.

hawkaa:
Excellent! Thank you!

PR #101, opened 2021-01-19
Serverless Plugin support two Cloudwatch log group subscription filters

zARODz11z:
### What does this PR do?
This PR modifies the Serverless plugin to support two Cloudwatch log group subscription filters. Additionally this PR adds a global const maxAllowableLogGroupSubscriptions variable that can be incremented in the future if Cloud watch adds support for 3 or more log group subscription filters.

### Motivation
Cloudwatch added support for two subscription filters per log group and the serverless plugin previously only supported one log group subscription filter. Adding this feature allows the plugin to successfully subscribe the forwarder to a log group in more scenarios.

### Testing Guidelines
I wrote 5 new unit tests in the forwarder.spec.ts file to cover the newly created test cases introduced by the support of 2 log group subscription filters. The tests are as follows:
1. Log group with no subscriptions -> Should subscribe
2. Log group with 1 subscription that is ours -> Should (re)subscribe
3. Log group with 1 subscription that is not ours -> Should subscribe
4. Log group with 2 subscriptions, 1 of which is ours -> Should (re)subscribe
5. Log group with 2 subscriptions, both are not ours -> **Should not subscribe**

Additionally I manually tested and deployed the 5 scenarios with mock log groups in the AWS console.

### Types of changes

- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


src/forwarder.ts
```diff
@@ -51,7 +51,7 @@ export async function addCloudWatchForwarderSubscriptions(service: Service, aws:
 
     const canSub = await canSubscribeLogGroup(aws, logGroupName, expectedSubName);
     if (!canSub) {
-      errors.push(`Subscription already exists for log group ${logGroupName}. Skipping subscribing Datadog forwarder.`);
+      errors.push(`Two non Datadog Subscriptions already exist for the log group ${logGroupName}. Skipping subscribing from Datadog forwarder.`);
```
nhinsch, line 54: Technically, we don't know for sure that there were exactly two non-Datadog subscriptions. While that would be the case for now, in the future AWS could add support for a third subscription (or more). Therefore, I would rephrase this to `Could not subscribe Datadog Forwarder due to too many existing subscription filter(s) for ${logGroupName}.`

src/forwarder.spec.ts
```diff
@@ -35,6 +36,59 @@ function awsMock(existingSubs: { [key: string]: any }, stackName?: string): Aws
   } as Aws;
 }
 
+describe("canSubscribeLogGroup",() => {
+  it("tells addCloudWatchForwarderSubscriptions function it can subscribe to a given log group with 0 existing subscription filters", async() => {
```
nhinsch, line 40: For all of these tests, I would remove the part where you say "tells addCloudWatchForwarderSubscriptions function...".

It's true that the `canSubscribeLogGroup()` function is used right now by the `addCloudWatchForwarderSubscriptions()` function, but this could be refactored or renamed in the future. So it would be better to really just focus in on what this function does in isolation. That's why we call them "unit tests" -- they test one unit (function) and that unit only.

So in this case I would phrase the tests as "Returns true if log group has 0 existing subscription filters", and the same pattern for the other tests. What we do after this function returns true or false is something that can be tested elsewhere -- these tests just focus on whether this one function returns true or false.

src/forwarder.spec.ts
```diff
@@ -35,6 +36,59 @@ function awsMock(existingSubs: { [key: string]: any }, stackName?: string): Aws
...
+
+   const aws = awsMock({});
+   const logGroupName: string = "/aws/lambda/serverless-plugin-test-dev-hello";
+   const expectedSubName: string = "serverless-plugin-test-dev-HelloLogGroupSubscription";
+
+   const canSubscribe = await canSubscribeLogGroup(aws,logGroupName,expectedSubName);
+   expect(canSubscribe).toBeTruthy();
```
nhinsch, line 47: Instead of `.toBeTruthy()` and `.toBeFalsy()` I would use `.toBe(true)` and `.toBe(false)`.

Truthy means that the variable will evaluate to true in an if statement. So the following values are all truthy: `true`, `754`, and `potato`. Likewise falsy could include `null` or `undefined`. We really want to check that the function actually returns `true`, not just something "truthy".

zARODz11z: That is great to know thanks!

nhinsch (approved):

PR #104, opened 2021-02-01
Serverless plugin support for Enhanced Plugins Object

zARODz11z:
### What does this PR do?
This PR changes the logic of how we decide whether the serverless-webpack plugin is present in a Serverless config file by adding support for a [enhanced plugins object](https://www.serverless.com/framework/docs/providers/aws/guide/plugins/). This enables the customer to link local custom plugins in their Serverless deployments along with externally installed Serverless plugins such as serverless webpack or serverless-plugin-datadog.

### Motivation
Previously the Datadog serverless plugin only supported a normal plugin array object but when a customer would include an [enhanced plugin object](https://www.serverless.com/framework/docs/providers/aws/guide/plugins/) the deployment would fail due to the logic in the hasWebpackPlugin method.

### Testing Guidelines
I added 5 unit tests for the hasWebpackPlugin method and manually tested through multiple deployments with different Serverless config settings.

### Additional Notes
On my machine the index.spec.ts unit test file is failing due to a unhandled promise rejection which should have nothing to do with the changes I added since it is occurring on both my master and development branch. I am also attempting to manually install the webpack plugin on a Serverless deployment as the final line of testing.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/env.spec.ts
+++ b/src/env.spec.ts
@@ -15,7 +15,7 @@ import {
 } from "./env";
 
 describe("hasWebpackPlugin", () => {
-  it("returns false when the serverless.yml plugin(s) object is not defined", () => {
+  it("returns false when the serverless.yml plugins object is not defined", () => {
     const service = {
       plugins: undefined,
     } as any;
@@ -23,7 +23,7 @@ describe("hasWebpackPlugin", () => {
     expect(result).toBe(false);
   });
 
-  it("returns false when the serverless.yml plugin(s) object does not define the serverless-webpack plugin", () => {
+  it("returns false when the serverless.yml plugins object does not define the serverless-webpack plugin", () => {
     const service = {
       plugins: ["serverless-plugin-datadog"],
     } as any;
@@ -31,7 +31,7 @@ describe("hasWebpackPlugin", () => {
     expect(result).toBe(false);
   });
 
-  it("returns true when the serverless.yml plugin(s) object does define the serverless-webpack plugin", () => {
+  it("returns true when the serverless.yml plugins object does define the serverless-webpack plugin", () => {
     const service = {
       plugins: ["serverless-plugin-datadog", "serverless-webpack"],
     } as any;
@@ -39,7 +39,7 @@ describe("hasWebpackPlugin", () => {
     expect(result).toBe(true);
   });
 
-  it("returns false when the serverless.yml enhanced plugin(s) object does not define the serverless-webpack plugin", () => {
+  it("returns false when the serverless.yml enhanced plugins object does not define the serverless-webpack plugin", () => {
     const service = {
       plugins: {
         localPath: "",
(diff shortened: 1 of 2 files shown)


src/env.ts
```diff
@@ -141,10 +141,25 @@ function getPropertyFromPath(obj: any, path: string[]) {
   return obj;
 }
 
+/*
+  ~check that plugins exists, if it does not then return false
+  ~decide if you have a normal array object or a enhanced object
```
zARODz11z, line 146: I left these comments here to make reviewing easier however I will remove them before merging.

nhinsch: In the future, no need to put these comments in for reviewing 🙂 ideally the code should be clear enough to understand on its own, as yours is! If you do want to provide more context for reviewing I would suggest putting it in the PR description instead of in the code itself.

src/env.ts
```diff
@@ -141,10 +141,25 @@ function getPropertyFromPath(obj: any, path: string[]) {
...
   const plugins: string[] | undefined = (service as any).plugins;
   if (plugins === undefined) {
     return false;
   }
-  return plugins.find((plugin) => plugin === "serverless-webpack") !== undefined;
+  if (Array.isArray(plugins)) {
+    // We have a normal plugin array
+    return plugins.find((plugin) => plugin === "serverless-webpack") !== undefined;
+  }
+  // We have a enhanced plugins object
```
nhinsch, line 159: Nit: a enhanced => an enhanced

src/env.ts
```diff
@@ -141,10 +141,25 @@ function getPropertyFromPath(obj: any, path: string[]) {
...
+*/
 export function hasWebpackPlugin(service: Service) {
   const plugins: string[] | undefined = (service as any).plugins;
   if (plugins === undefined) {
     return false;
   }
-  return plugins.find((plugin) => plugin === "serverless-webpack") !== undefined;
+  if (Array.isArray(plugins)) {
+    // We have a normal plugin array
+    return plugins.find((plugin) => plugin === "serverless-webpack") !== undefined;
```
nhinsch, line 157: When you have a certain string that you look for in multiple places, I would recommend pulling that out into a constant. I.e.

```
WEBPACK_PLUGIN_NAME = "serverless-webpack"
```

Then you can use `WEBPACK_PLUGIN_NAME` in the code. It makes it just a little bit more clear what the significance of that particular string is.

src/env.spec.ts
```diff
@@ -6,7 +6,61 @@
...
+import {
+  getConfig,
+  defaultConfiguration,
+  setEnvConfiguration,
+  forceExcludeDepsFromWebpack,
+  hasWebpackPlugin,
+} from "./env";
+
+describe("hasWebpackPlugin", () => {
+  it("returns false when the serverless.yml plugin(s) object is not defined", () => {
```
nhinsch, line 18: Nit: I think the object is always called `plugins` so I would suggest `plugin(s)` => `plugins`

nhinsch: (same in the other tests)

nhinsch (approved):
Awesome! 🤩 I just made a few small suggestions!

PR #105, opened 2021-02-02
Fix error message to reflect incorrect function ARN 

hjiang27:
### What does this PR do?
Throws DatadogForwarderNotFoundError when an invalid forwarder ARN is used. Added a unit test to check that DatadogForwaderNotFoundError is thrown. 

Screen Shot 2021-01-29 at 7 56 23 AM

Screen Shot 2021-01-29 at 7 07 08 AM

### Motivation
The original error thrown ("An error occurred: GetaccountapiLogGroupSubscription - Could not execute the lambda function. Make sure you have given CloudWatch Logs permission to execute your function.") was misleading to users and did not specify that the subscription was unsuccessful due to an incorrect function ARN. This PR is meant to clarify the error. 

### Testing Guidelines
Tested manually with an incorrect function ARN.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply


src/forwarder.ts
```diff
@@ -28,12 +39,25 @@ interface DescribeSubscriptionFiltersResponse {
 function isLogGroup(value: any): value is LogGroupResource {
   return value.Type === logGroupKey;
 }
+/**
+ * Validates whether ARN exists in the account
+ * @param aws Serverless framework provided AWS client
+ * @param functionArn The ARN to be validated
+ */
+async function doesArnExist(aws: Aws, functionArn: string) {
```
nhinsch, line 47: I would rename this function to `doesFunctionExist` -- ARNs are used for lots of things besides functions, and this code specifically looks for a Lambda function.

src/forwarder.ts
```diff
@@ -28,12 +39,25 @@ interface DescribeSubscriptionFiltersResponse {
...
+  try {
+    const result = await aws.request("Lambda", "getFunction", { FunctionName: functionArn });
+  } catch (err) {
+    throw new DatadogForwarderNotFoundError(`Could not perform GetFunction on ${functionArn}.`);
```
nhinsch, line 51: I would not throw `DatadogForwarderNotFoundError` in a function that checks for the existence of a Lambda that may or may not be the Forwarder. What if we called this function on a non-Forwarder lambda? If we didn't find it, we wouldn't want the error to say anything about the Forwarder.

A couple of ideas:
- You could rename this function so that it is clear that it's specifically looking for a Forwarder, not any old Lambda.
- You could leave the function more general, but then move the error handling code outside of this function into a context where we know we're talking about the Forwarder.

src/index.spec.ts
```diff
@@ -18,7 +18,12 @@ const SEM_VER_REGEX = /^v(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\
 function awsMock(): Aws {
   return {
     getStage: () => "dev",
-    request: (service, method, params: any) => Promise.reject("Log group doesn't exist"),
+    request: (service, method, params: any) => {
+      if (method == "getFunction") {
+        return Promise.resolve();
+      }
+      return Promise.reject("Log group doesn't exist");
```
nhinsch, line 25: It seems a bit odd to have the Promise reject with "Log group doesn't exist" for lots of methods that don't have anything to do with log groups. Maybe instead of having `getFunction` be the special case, you can just have the rejection only occur when looking for a log group, and make the everything else resolve?

src/forwarder.ts
```diff
@@ -6,6 +6,17 @@ const logGroupKey = "AWS::Logs::LogGroup";
 const logGroupSubscriptionKey = "AWS::Logs::SubscriptionFilter";
 const maxAllowableLogGroupSubscriptions: number = 2;
 
+/**
+ * Error to be thrown when function ARN cannot be found.
+ */
```
nhinsch, line 11: Nit: I'm not sure you need this comment here -- it's pretty obvious what the error does even without the comment. And that's a good thing! 🙂

src/forwarder.spec.ts
```diff
@@ -285,4 +291,44 @@ describe("addCloudWatchForwarderSubscriptions", () => {
       }
     `);
   });
+  it("throws DatadogForwarderNotFoundError", async () => {
```
nhinsch, line 294: I would update the name of the test to say WHEN/WHY it throws DatadogForwarderNotFoundError. I.e. "it throws DatadogForwarderNotFoundError when ..."

src/forwarder.spec.ts
```diff
@@ -20,11 +20,17 @@ function serviceWithResources(resources?: Record<string, any>, serviceName = "my
...
     getStage: () => "dev",
     request: (service, method, params: any) => {
+      if (doesAlwaysReject) {
+        return Promise.reject("Not found.");
+      }
       const logGroupName = params.logGroupName;
+      if (method == "getFunction") {
+        return Promise.resolve();
+      }
       if (existingSubs[logGroupName]) {
```
nhinsch, line 34: It looks like this logic here around existing subs is intended for one specific method (getting subscription filters?), I would wrap it in an `if` statement so that it's scoped to that method only.

src/forwarder.ts
```diff
@@ -28,12 +36,25 @@ interface DescribeSubscriptionFiltersResponse {
 function isLogGroup(value: any): value is LogGroupResource {
   return value.Type === logGroupKey;
 }
+/**
+ * Validates whether ARN exists in the account
```
nhinsch, line 40: Don't forget to update the comment too 🙂 you are validating whether a Lambda function exists, not just any ARN

nhinsch (approved):
Awesome job! Just a couple minor suggestions before merging! 🔥

src/forwarder.ts
nhinsch: And specifically a Forwarder!

PR #106, opened 2021-02-03
Update SLS plugin to support"forwarder" or "forwarderArn" parameters

zARODz11z:
### What does this PR do?
This PR updates the Serverless Plugin "forwarder" parameter to "forwarderArn" to match the configuration settings of the Datadog CloudFormation Macro while still keeping support for the "forwarder" parameter in efforts of not breaking previously instrumented Serverless functions. This PR also updates the Serverless Plugin README to document the capability to use either "forwarder" or "forwarderArn" parameters.

### Motivation
My motivation for this PR is that the [Datadog Serverless Plugin](https://github.com/DataDog/serverless-plugin-datadog) used the "forwarder" parameter while the [Datadog CloudFormation Macro](https://github.com/DataDog/datadog-cloudformation-macro/tree/master/serverless) used the "forwarderArn" parameter, "forwarder" and "forwarderArn" are synonymous as they are ultimately configured to take in the Datadog forwarder function arn. Updating "forwarder" to "forwarderArn" keeps the convention consistent between the SLS plugin and CFM macro and hopefully eliminates any previous confusion.

### Testing Guidelines
I created unit tests in the index.spec.ts file covering the use of "forwarderArn" and "forwarder" parameters in the Serverless.yml file which confirmed expected results through the presence of a created subscription filter. I also manually configured a test Serverless function with the use of "forwarderArn" and "forwarder" and confirmed the results in the AWS console (CloudWatch Log groups).

### Additional Notes
I had a local issue running the index.spec.ts file but this problem was persistent on the master branch as well as my development branch so it is most likely a machine problem.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/index.ts
```diff
@@ -95,10 +95,30 @@ module.exports = class ServerlessPlugin {
...
+    const forwarderArn: string | undefined = config.forwarderArn;
+    const forwarder: string | undefined = config.forwarder;
+
+    let functionArn;
     if (config.enabled === false) return;
-    if (config.forwarder) {
+
+    if (forwarderArn && forwarder) {
+      console.error(
+        "Error: Could not add CloudWatch forwarder subscriptions because both 'forwarderArn' and 'forwarder' parameters are set. 'forwarderArn' and 'forwarder' are equivalent, please only use 'forwarderArn' or 'forwarder'.",
```
nhinsch, line 106: Likewise here, I would encourage customers to specifically use `forwarderArn` if they have set both.

src/index.spec.ts
```diff
@@ -252,6 +287,73 @@ describe("ServerlessPlugin", () => {
...
+        },
+      };
+      const plugin = new ServerlessPlugin(serverless, {});
+      await plugin.hooks["after:package:createDeploymentArtifacts"]();
+      expect(serverless.service.provider.compiledCloudFormationTemplate.Resources).not.toHaveProperty(
+        "FirstGroupSubscription",
+      );
+    });
+
+    it("does not add subscription filters when neither fowarderArn or forwarder are set", async () => {
```
nhinsch, line 325: Nit: neither ... or -> neither ... nor

nhinsch: Also, fowarderArn -> forwarderArn

src/index.spec.ts
```diff
@@ -252,6 +287,73 @@ describe("ServerlessPlugin", () => {
       );
     });
 
+    it("does not add subscription filters when fowarderArn and forwarder is set", async () => {
```
nhinsch, line 290: Nit: is set -> are set

src/env.ts
```diff
@@ -28,9 +28,9 @@ export interface Configuration {
   // Enable tracing on Lambda function using dd-trace, datadog's APM library.
   enableDDTracing: boolean;
 
-  // When set, the plugin will subscribe the lambdas to the forwarder with the given arn.
+  // When either are set, the plugin will subscribe the lambdas to the forwarder with the given arn.
```
nhinsch, line 31: Nit: either are -> either is

src/index.ts
```diff
@@ -95,10 +95,30 @@ module.exports = class ServerlessPlugin {
 
   private async afterPackageFunction() {
     const config = getConfig(this.serverless.service);
+    const forwarderArn: string | undefined = config.forwarderArn;
+    const forwarder: string | undefined = config.forwarder;
+
+    let functionArn;
```
nhinsch, line 101: I think the name of this variable could potentially be a bit confusing because it seems to be the ARN of the Forwarder, not of the function being deployed by the plugin.

src/index.ts
```diff
@@ -95,10 +95,30 @@ module.exports = class ServerlessPlugin {
...
+    if (forwarderArn && forwarder) {
+      console.error(
+        "Error: Could not add CloudWatch forwarder subscriptions because both 'forwarderArn' and 'forwarder' parameters are set. 'forwarderArn' and 'forwarder' are equivalent, please only use 'forwarderArn' or 'forwarder'.",
+      );
+    } else if (forwarderArn !== undefined && forwarder === undefined) {
+      functionArn = forwarderArn;
+    } else if (forwarder !== undefined && forwarderArn === undefined) {
+      functionArn = forwarder;
+    } else {
+      console.error(
```
nhinsch, line 113: I think it might be a better experience for customers who set both `forwarder` and `forwarderArn` for us to just favor `forwarderArn` and perhaps display a warning, rather than potentially deleting their existing log group subscriptions. Alternatively, maybe there is a way to make the deployment completely fail until the customer corrects the config? Either way, I don't think we should ever delete log group subscriptions because the customer adds a redundant ARN to their config.

src/index.ts
zARODz11z: I am a bit confused with the comment, "let functionArn" is supposed to represent the Datadog Forwarder Arn that is passed into the addCloudWatchForwarderSubscriptions function, I changed the "functionArn" variable to "datadogForwarderArn" to be more explicit.

src/index.ts
zARODz11z: Hmm yeah this crossed my mind also but I couldn't find a way of terminating the deployment, I am going to investigate how I can terminate it with an error message telling the customer to correct the config.

zARODz11z: I edited the code so that if both forwarder and forwarderArn are set then a error will be thrown telling the customer to only use forwarderArn. I tested this with an existing forwarder subscription and the thrown error did not cause the subscription to be deleted, instead just the deployment failed. 
This is also the case of if no neither the forwarder nor the forwarderArn is provided. Only the current deployment will fail, no subscriptions will be lost anymore.

README.md
```diff
@@ -31,8 +31,8 @@ To further configure your plugin, use the following custom parameters in your `s
...
-| `forwarder`          | Setting this parameter subscribes the Lambda functions' CloudWatch log groups to the given Datadog forwarder Lambda function. Required when `enableDDTracing` is set to `true`.                                                                                                                                                                                                                 |
+| `enableDDTracing`    | Enable Datadog tracing on the Lambda function. Defaults to `true`. When enabled, it's required to set either the `forwarderArn` or `forwarder` parameter there is no difference in which you choose but please only use one .                                                                                                                                                                                                                                                                         |
```
nhinsch, line 34: I would remove the bit on the end about the forwarder parameter and just mention the forwarderArn parameter.

src/index.ts
```diff
@@ -95,10 +95,29 @@ module.exports = class ServerlessPlugin {
...
+    if (forwarderArn && forwarder) {
+      throw new Error(
+        "Error: Both 'forwarderArn' and 'forwarder' parameters are set. Please only use the 'forwarderArn' parameter.",
+      );
+    } else if (forwarderArn !== undefined && forwarder === undefined) {
+      datadogForwarderArn = forwarderArn;
+    } else if (forwarder !== undefined && forwarderArn === undefined) {
+      datadogForwarderArn = forwarder;
+    } else {
+      throw new Error(
```
nhinsch, line 113: Does this mean the `forwarder`/`forwarderArn` is no longer optional? We want to still be able to deploy without these parameters (we just won't create subscription filters)

zARODz11z: Deploying without forwarder/forwarderArn is still optionally, a warning message appears but deployment still continues

src/index.spec.ts
```diff
@@ -224,6 +224,40 @@ describe("ServerlessPlugin", () => {
       mock.restore();
     });
     it("adds subscription filters when fowarderArn is set", async () => {
```
nhinsch, line 226: fowader -> forwarder

src/index.ts
```diff
@@ -95,13 +95,29 @@ module.exports = class ServerlessPlugin {
...
+
+    if (datadogForwarderArn) {
       const aws = this.serverless.getProvider("aws");
-      const errors = await addCloudWatchForwarderSubscriptions(this.serverless.service, aws, config.forwarder);
+      const errors = await addCloudWatchForwarderSubscriptions(this.serverless.service, aws, datadogForwarderArn);
       for (const error of errors) {
         this.serverless.cli.log(error);
       }
+    } else {
+      this.serverless.cli.log("Warning: 'forwarderArn' is not set, skipping CloudWatch Forwarder Subscriptions.");
```
nhinsch, line 120: I don't think we need a warning here. It's normal to not provide the forwarderArn if you don't want to. There are other ways to set up the subscriptions besides the plugin, so it's not totally unexpected or something we would need to warn customers about.

nhinsch (approved):
Looks great, just a couple minor suggestions.

PR #115, opened 2021-03-01
Datadog Extension to Serverless Plugin

zARODz11z:
### What does this PR do?
This PR adds support for the Datadog Extension Layer in the Datadog Serverless Plugin.

### Motivation
In the future the Datadog Extension layer may be a simpler solution to instrumenting Serverless functions as opposed to the Datadog Forwarder. I believe the customer should be able to choose between the two options.

### Testing Guidelines
I created unit tests to cover the scenarios I could think of while also manually testing by linking my development copy of the plugin to a Serverless framework application where I manually modified the YML configuration and deployed repeatedly. From there I confirmed results in the AWS console then also invoking the functions and viewing the results in the Datadog app.

### Additional Notes
I am not sure if enforcing a API key when using the extension is the right thing. On one hand it would simplify a lot of problems if a customer was enforced by the plugin to provide an api key when using the extension, while on the other hand what if the customer wishes to add their API key through the aws console instead of the serverless application?

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


README.md
```diff
@@ -24,15 +24,16 @@ To further configure your plugin, use the following custom parameters in your `s
...
+| `apiKMSKey`          | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false` or `enableDDExtension` is `true`, and you are using KMS encryption.                                                                                                                                                                                                                                             |
+| `addLayers`          | Whether to install the Datadog Lambda library or Datadog Lambda Extension as a layer. Defaults to `true`. Set to `false` when you plan to package the Datadog Lambda library to your function's deployment package on your own so that you can install a specific version of the Datadog Lambda library ([Python][4] or [Node.js][5]). |
```
nhinsch, line 31: Does setting `addLayers` have an effect on how we install the extension? I thought this parameter was just for the regular layer but I might be wrong.

README.md
```diff
@@ -24,15 +24,16 @@ To further configure your plugin, use the following custom parameters in your `s
...
-| `forwarderArn`          | Setting this parameter subscribes the Lambda functions' CloudWatch log groups to the given Datadog forwarder Lambda function. Required when `enableDDTracing` is set to `true`.                                                                                                                                                                                                                 |
+| `enableDDExtension`    | Enable Datadog Lambda Extension Layer. Defaults to `false`. When enabled, it's required to set the `apiKey` (or `apiKMSKey`) parameter. Do not set `forwarderArn` or `flushMetricsToLogs` while using `enableDDExtension`.                                                                                                                                                                                                                                                                         |
```
nhinsch, line 35: Nit: I would call this `addExtension` to be consistent with `addLayers`

src/layer.ts
```diff
@@ -74,17 +74,33 @@ export function applyLayers(region: string, handlers: FunctionInfo[], layers: La
...
-      const currentLayers = getLayers(handler);
-      if (!new Set(currentLayers).has(layerARN)) {
-        currentLayers.push(layerARN);
-      }
+    const lambdaLayerARN = runtime !== undefined ? regionRuntimes[runtime] : undefined;
+    let extensionLayerARN: string | undefined;
+    if (applyExtensionLayer) extensionLayerARN = regionRuntimes[extensionLayerKey];
+    let currentLayers = getLayers(handler);
+
+    if (lambdaLayerARN && extensionLayerARN) {
```
nhinsch, line 82: Do you need three cases here? I.e. do you need to check `lambdaLayerARN && extensionLayerARN`, `lambdaLayerARN`, and `extensionLayerARN` separately? Can you call `setLayers` more than once? If so, then I think you might only need to check for the `lambdaLayerARN` and `extensionLayerARN` separately.

README.md
zARODz11z: As of right now if addLayers is false and addExtension is true then the extension will not be added because the addLayers parameter acts like a on/off switch for all layers although after thinking about it I think addLayers should just be regular layers and addExtension is its separate thing.

zARODz11z: Per our slack convo, addLayers and addExtension are now independent of another. The addLayers parameter strictly adds the lambda library as a layer while addExtension strictly adds the lambda extension as a layer.

src/layer.ts
zARODz11z: I refactored the logic

src/layers.json
```diff
@@ -7,7 +7,8 @@
       "python2.7": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Python27:29",
       "python3.6": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Python36:29",
       "python3.7": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Python37:29",
-      "python3.8": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Python38:29"
+      "python3.8": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Python38:29",
+      "extension": "arn:aws:lambda:af-south-1:464622532012:layer:Datadog-Extension:5"
```
zARODz11z, line 11: The modification to the generate_layers.sh script generated the addition of the extension arns in the layers.json file however the script was not run to generate the extension arns for the layer-gov.json file (for gov cloud).

zARODz11z: I was unable to generate those arns via the script because I do not have gov cloud access.

README.md
```diff
@@ -24,15 +24,16 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `addLayers`          | Whether to install the Datadog Lambda library as a layer. Defaults to `true`. Set to `false` when you plan to package the Datadog Lambda library to your function's deployment package on your own so that you can install a specific version of the Datadog Lambda library ([Python][4] or [Node.js][5]). |
+| `addExtension`    | Whether to install the Datadog Lambda Extension as a layer. Defaults to `false`. When enabled, it's required to set the `apiKey` (or `apiKMSKey`) parameter. While using `addExtension` set `flushMetricsToLogs` to `false` and leave out `forwarderArn`. Learn more about the Lambda Extension [here][8].                                                                                                                                                                                                                                                                      |
```
nhinsch, line 32: I would add the note that the Lambda Extension is in preview (like we have in the CDK repo)

README.md
```diff
@@ -24,15 +24,16 @@ To further configure your plugin, use the following custom parameters in your `s
...
-| `forwarderArn`          | Setting this parameter subscribes the Lambda functions' CloudWatch log groups to the given Datadog forwarder Lambda function. Required when `enableDDTracing` is set to `true`.                                                                                                                                                                                                                 |
+| `forwarderArn`          | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true`. Do not set `addExtension` to `true` when using `forwarderArn`.                                                                                                                                                                                                                 |
```
nhinsch, line 36: 👌

src/env.spec.ts
```diff
@@ -85,6 +85,32 @@ describe("getConfig", () => {
...
       enableXrayTracing: false,
       enableDDTracing: true,
+      addExtension: false,
+      enableTags: true,
+      injectLogContext: true,
+      exclude: [],
+    });
+  });
+
+  it("gets a mixed configuration when addExtension is set", () => {
```
nhinsch, line 95: What do you mean here by "gets a mixed configuration"?

src/env.ts
```diff
@@ -11,7 +11,7 @@ import Service from "serverless/classes/Service";
 export interface Configuration {
   // Whether Datadog is enabled. Defaults to true.
   enabled?: boolean;
-  // Whether to add the lambda layers, or expect the user's to bring their own
+  // Whether to add the lambda library layers, or expect the user's to bring their own
```
nhinsch, line 14: Good clarification

src/index.ts
```diff
@@ -66,20 +66,29 @@ module.exports = class ServerlessPlugin {
...
     this.serverless.cli.log("Auto instrumenting functions with Datadog");
+    validateConfiguration(config);
     setEnvConfiguration(config, this.serverless.service);
 
     const defaultRuntime = this.serverless.service.provider.runtime;
     const handlers = findHandlers(this.serverless.service, config.exclude, defaultRuntime);
+    const allLayers = { regions: { ...layers.regions, ...govLayers.regions } };
     if (config.addLayers) {
-      this.serverless.cli.log("Adding Lambda Layers to functions");
+      this.serverless.cli.log("Adding Lambda library Layers to functions");
```
nhinsch, line 76: Super nit: I think `Library` should be capitalized here as well (`Lambda Library Layers`) -- technically the library is called the Datadog Lambda Library.

src/layers.json
nhinsch: I think that should be fine as long as the next person who releases the plugin has GovCloud access. Are you confident that the script will work for GovCloud for someone who does have access?

nhinsch (approved):
Looks good! Just a few minor suggestions!

src/layers.json
zARODz11z: The publish_prod.sh script checks if the user wants to update the layers while publishing a new version to production [here](https://github.com/DataDog/serverless-plugin-datadog/blob/41a10c6a67039218969c68bd56ebadae167825a8/scripts/publish_prod.sh#L50). Inside there the user needs to have gov cloud and prod-engineering access to run the generate_layers_json script. I am confident that if someone who has gov cloud access ran the publish script while also wishing to update the layers then both regular and gov cloud arns should be generated.

src/env.spec.ts
zARODz11z: I took the "gets a mixed configuration" from the test above but from what I understand is that when the customer provides actual parameters under the custom section then we get a config that is mixed between defaults and stuff that the customer explicitly defined. 
So its not entirely the default or custom but its a mix between the both.

src/layers.json
zARODz11z: There may be a limbo period where gov cloud customers wishing to use the extension may not be able to until someone updates those layers.

src/env.spec.ts
nhinsch: Hmm I think "mixed configuration" is a bit confusing, I would consider rewording this test case. What, in your words, is this test case trying to test?

src/layers.json
nhinsch: I don't think there would be a gap because the changes in this PR won't be available until someone releases a new version of this plugin, which would entail updating the layers, right?

PR #117, opened 2021-03-09
Ignore flushMetricsToLogs when using Extension

zARODz11z:
### What does this PR do?
This PR simplifies the process for adding the Datadog Lambda Extension to via the SLS Plugin. Before the yml to add the extension looked like this: 
```
custom:
  datadog:
    flushMetricsToLogs: false
    addExtension: true
    apiKey: "1234"
```
If the user did not explicitly set flushMetricsToLogs to false then their deployment would fail with an error saying "do not set flushMetricsToLogs while using addExtension". This PR adds logic so that if addExtension is true then the environment variable assoiciated with flushMetricsToLogs (DD_FLUSH_TO_LOG) will not be added in the lambda config instead of throwing an error. This simplifies the necessary yml to be: 
```
custom:
  datadog:
    addExtension: true
    apiKey: "1234"
```

### Testing
I added a unit test to make sure DD_FLUSH_TO_LOG would not be added when addExtension was true while existing unit tests covered the case on if addExtension was false, in that case DD_FLUSH_TO_LOG would be set accordingly. Afterwards I manually tested these scenarios via a SLS framework demo and asserted results in AWS.
### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


nhinsch (approved):
Looks great! 😎 🌴

nhinsch:
I would rename the PR though, not sure what "Support Patch #1" means. The PR name will go into the commit log so it's good to have a descriptive name.

PR #120, opened 2021-03-29
fixes #119 - clarifies when forwarderArn actually is required

### LINKED ISSUE #119: Clarify Behavior around `forwarderArn` option
## Expected Behavior
[The documentation](https://www.serverless.com/plugins/serverless-plugin-datadog) for the `forwarderArn` states: 

> Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when enableDDTracing is set to true. Do not set addExtension to true when using forwarderArn.

Because it its `Required`, we'd expect an error to be thrown. **WE ACTUALLY DO NOT WANT THIS BEHAVIOR**

## Actual Behavior

When something else has added the datadog forwarder - such as the Datadog AWS Integration - tracing works as expected. **WE ACTUALLY DO LIKE THIS BEHAVIOR**

We would like this to continue to be the case and not have the `forwarder` or `forwarderArn` be required. Right now, this works as desired but I'd like to have the docs changed to reflect that you just need SOMETHING to create the forwarder and not require it in serverless.yaml. 

Our use case: 

* We have dozens, maybe over a hundred lambdas - we have even more log groups
* Half of the lambdas are in serverless, the other half are in terraform, we also have non-lambda things generating logs
* We use the DD aws integtration to automatically subscribe our log groups to the forwarder
* We'd prefer to allow the DD AWS integration to handle subscription - if it is off, devs have to rememeber to add it, if its on but devs add this, there is a double subscription

**I am happy to open an MR to do this** but before I do, I just want to make sure there isn't some other reason to leave it as is. What we want to avoid is someone seeing that blurb, then seeing that serverless isn't enforcing that behavior, then updating serverless to enforce that behavior.

## Steps to Reproduce the Problem

  1. In serverless, create a lambda w/o tracing or a sub filter
  2. Outside of serverless, create a subscription filter to a datadog forwarder
  3. In serverless, specific `enableDDTracing: true`, do not specifiy `forwarder` or `forwarderArn`, trace any function
  4. Deploy and test tracing -> this will deploy and trace w/o error

## Specifications

  - Serverless Framework version: 1.78.1
  - Datadog Serverless Plugin version: 2.17.0
  - Lambda function runtime (Python 3.7, Node 10, etc.): Python 3.8

jwyoungpm:
### What does this PR do?

Changes the documentation for the `forwarderArn` option. It is currently phrased as such: 

> Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when enableDDTracing is set to true. Do not set addExtension to true when using forwarderArn.

However, serverless does _not_ throw an error when this is missing if `enableDDTracing` is set to `true` - I believe the current behavior is actually correct, so this updates the doc to reflect this. 

### Motivation

Our use case demonstrates why this should not be required: 

* We are in aws, we use the AWS Integration provided by DataDog to add subscriptions automagically
* We have a number of services outside of lambda that would benefit from this auto subscription
* We want to provide direction to our developers that they should let the integration take care of this in every situation, this way we get consistent behavior around subscriptions

The reason for the vernacular change is that I do NOT want someone to see the current phrasing and think:

> No error is thrown when `forwarderArn` is missing if `enableDDTracing` is `true`, that must be a bug that I should fix

This also helps to solve some confusion for those that ARE using the DD aws integration to create such subscriptions.

### Testing Guidelines

N/A - no code changes

### Additional Notes

This is my first MR, I did read contrib guidelines but let me know if there are any issues. I assume a change to the README would be reflected in the actual docs here, let me know if that is not the case: https://www.serverless.com/plugins/serverless-plugin-datadog

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [X] This PR's description is comprehensive


README.md
```diff
@@ -33,7 +33,7 @@ To further configure your plugin, use the following custom parameters in your `s
...
-| `forwarderArn`          | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true`. Do not set `addExtension` to `true` when using `forwarderArn`.                                                                                                                                                                                                                 |
+| `forwarderArn`          | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true` unless the subscription is otherwise applied. If a subscription is applied via an integration such as DataDogs AWS Integration, this is not required. Do not set `addExtension` to `true` when using `forwarderArn`.                                                                                                                                                                                                                 |
```
zARODz11z, line 36: Nit: I think Datadogs -> Datadog's

README.md
```diff
@@ -33,7 +33,7 @@ To further configure your plugin, use the following custom parameters in your `s
...
-| `forwarderArn`          | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true`. Do not set `addExtension` to `true` when using `forwarderArn`.                                                                                                                                                                                                                 |
+| `forwarderArn`          | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true` unless the subscription is otherwise applied. If a subscription is applied via an integration such as DataDog's AWS Integration, this is not required. Do not set `addExtension` to `true` when using `forwarderArn`.                                                                                                                                                                                                                 |
```
zARODz11z, line 36: Perhaps we should be a bit more specific about the `subscription` and what is `not required`. I think the sentence starting with "If a subscription is ..."
could be rephrased to "For example if a Datadog Forwarder subscription is applied via Datadog's AWS Integration, then `forwarderArn` is not required."

zARODz11z (approved):
Thank you for catching this, we appreciate your contribution!

PR #121, opened 2021-04-12
Add snapshot Integration test(s)

zARODz11z:
### What does this PR do?
Adds snapshot integration testing to the Serverless Plugin

### Motivation
Having this sort of integration test can give us more confidence when merging new changes.

### Testing Guidelines
I manually ran the test script with various scenarios testing out the codepaths.

### Additional Notes
I am storing the snapshots that we assert against as the `correct_extension_snapshot` and `correct_forwarder_snapshot`. Some of the lines in the CloudFormation template seem to be generated randomly so I had to use a `Regex` and `grep` to remove them before feeding them into the `diff` command.

I had 2 different yml files, the `serverless-extension.yml` and `serverless-forwarder.yml` due to the nature of the SLS plugin the Datadog Forwarder and Lambda Extension Layer cannot be configured at the same time and I felt it was important to test both of them.

I added functions for every supported JS and Python runtimes so that we can have the most coverage on the Plugin adding the correct layers to the functions. It may have increased the size of the CloudFormation template files however I think it's a good idea to test the addition of all the layers for different runtimes.

In favor of not making any requests to AWS to validate the forwarderArn I added logic to make `arn:aws:lambda:us-east-1:000000000000:function:datadog-forwarder` be the forwarderArn that is used for Integration testing (Check out the change in the `forwarder.ts` file. This allows the integration tests to be more consistent and not need AWS credentials.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


integration_tests/js_handler.js
```diff
@@ -0,0 +1,18 @@
...
+      {
+        message: 'Go Serverless v1.0! Your function executed successfully!',
+        input: event,
+      },
+      null,
+      2
+    ),
+  };
+
+  // Use this code if you don't use the http event with the LAMBDA-PROXY integration
```
DarcyRaynerDD, line 16: I'm guessing this file is autogenerated, but we can probably remove this comment

integration_tests/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "integration_tests",
+  "description": "",
+  "version": "0.1.0",
+  "dependencies": {},
+  "devDependencies": {
+    "serverless-plugin-datadog": "^2.18.0"
```
DarcyRaynerDD, line 7: Is this reference to the plugin necessary?

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,79 @@
...
+# To regenerate snapshots:
+#   UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests
+
+set -e
+
+script_path=${BASH_SOURCE[0]}
+scripts_dir=$(dirname $script_path)
+repo_dir=$(dirname $scripts_dir)
+root_dir=$(pwd)
+#/Users/andrew.rodriguez/go/src/github.com/DataDog/serverless-plugin-datadog
```
DarcyRaynerDD, line 15: We can delete this comment

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,79 @@
...
+
+cd $integration_tests_dir
+yarn
+
+# Add local build to node_modules so `serverless-forwarder.yml` and `serverless-extension.yml` also have access to local build.
+cd $root_dir
+yarn
+yarn build
+rm -rf "$integration_tests/node_modules"
+mkdir -p "$integration_tests_dir/node_modules/serverless-plugin-datadog"
```
DarcyRaynerDD, line 35: So, a trick you can do with the serverless plugins, is to put them into the [serverless plugins](https://www.serverless.com/framework/docs/providers/aws/guide/plugins#service-local-plugin) directory, instead of node_modules. Or altenatively, you can might be able to change 'serverless-plugin-datadog' to a local path, eg '../dist'.

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,79 @@
...
+cp -r dist "$integration_tests_dir/node_modules/serverless-plugin-datadog"
+
+cd $integration_tests_dir
+echo "Running 'sls package' with 'serverless-forwarder.yml'"
```
DarcyRaynerDD, line 39: There is a bit of duplication here. It would be nice to be able to easily add more tests/cases without duplicating code. Maybe we could write a loop, and have the test names in an array.

integration_tests/yarn.lock
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
DarcyRaynerDD, line 1: We can also delete this file, if we don't need the package.json

src/forwarder.ts
```diff
@@ -49,6 +49,14 @@ function isLogGroup(value: any): value is LogGroupResource {
  * @param functionArn The forwarder ARN to be validated
  */
 async function validateForwarderArn(aws: Aws, functionArn: CloudFormationObjectArn | string) {
+  if (functionArn === "arn:aws:lambda:us-east-1:000000000000:function:datadog-forwarder") {
```
DarcyRaynerDD, line 52: From the perspective of the code calling `validateForwarderArn`, this is a hidden case which breaks the expectation of what the function does. It's better to move the logic to skip validation to where 'validateForwarderArn' is called instead. The other issue is the arn is a magic value. An alternative might be to add another config value, eg `validateForwarder`,  default it to true and check it before calling `validateForwarderArn`. Then we set it to false for the integration tests.

DarcyRaynerDD (changes requested):

integration_tests/package.json
zARODz11z: It is not, we are using the local version of the plugin anyways so I removed this dependency.

scripts/run_integration_tests.sh
zARODz11z: Yup I removed the logic of adding the local plugin to the node modules and instead just referenced the local plugin in the yml files.

src/forwarder.ts
zARODz11z: I introduced a validateForwarder parameter, let me know how that looks

DarcyRaynerDD (commented):
This looks great Andrew. Last thing, can we add a github action to call the integration tests, and block if they don't match?

zARODz11z:
@DarcyRaynerDD I was planning on splitting this into 2 PR's but i'm glad I didn't because I was able to do some refactoring on the integration test script. I also changed the `validateForwarder` param to `integrationTesting` because in addition to bypassing the validation of the forwarder ARN during integration tests it turns out that I would also have to bypass the adding of Datadog Monitor output links to the CloudFormation template because it was throwing the diffs off. This should keep the integration testing consistent.  Some of the files that you reviewed may have changed, if you could take another look that would be great.

zARODz11z:
I'm pretty sure this goes without saying but I would just like to point it out that these snapshot integration tests will only fail if the user made a certain type of change to the plugin which would ultimately change the CloudFormation template. So this will not catch every change to the plugin because not everything changes the CFN template. On the bright side these snapshot integration tests should ensure that the primary functions of the plugin still works like adding Lambda layers, Forwarder subscriptions, Extension layer, and other products of the plugin that show up in the CFN template.

DarcyRaynerDD (approved):

PR #122, opened 2021-04-16
Update snapshots during publish to prod

zARODz11z:
### What does this PR do?
* This PR invokes `UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh` in the `publish_prod.sh` script to keep snapshots updated every-time there is a new release. 
* Added some minor edits to the `run_integration_tests.sh` script so that it can be called from the `publish_prod.sh` script and commit the changes correctly. 
* Added a check to make sure the `publish_prod.sh` script is called from the root directory in hopes of avoiding script failures when calling other scripts. If someone would run the script from the scripts directory like this `./publish_prod.sh` instead of from the root like this `./scripts/publish_prod.sh` then I believe [this](https://github.com/DataDog/serverless-plugin-datadog/blob/04b5ffc4d9670aa60a37e41b8f97526f32c10ffb/scripts/publish_prod.sh#L57) call would fail because it wouldn't be able to find the `scripts` directory if it was already in it.
* Finally I updated the snapshots which contained the new layer versions and SLS Plugin version

### Motivation
Keep snapshots updated when a release is made. The last [v.2.20.0](https://github.com/DataDog/serverless-plugin-datadog/commit/04b5ffc4d9670aa60a37e41b8f97526f32c10ffb) commit failed the integration test workflow because the snapshots were not updated so I think updating the snapshots should be apart of the release process in hopes of keeping things automated and in sync.

### Additional Notes
Some of the changes in the CloudFormation template are random like the end of identifiers (I exclude those during our integration tests with a regex) but the important things are the updated layers and SLS plugin version.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/integration_tests/correct_extension_snapshot.json
+++ b/integration_tests/correct_extension_snapshot.json
@@ -197,7 +197,7 @@
           "S3Bucket": {
             "Ref": "ServerlessDeploymentBucket"
           },
-          "S3Key": "serverless/dd-sls-plugin-integration-test/dev/1618443590964-2021-04-14T23:39:50.964Z/dd-sls-plugin-integration-test.zip"
+          "S3Key": "serverless/dd-sls-plugin-integration-test/dev/1618590312126-2021-04-16T16:25:12.126Z/dd-sls-plugin-integration-test.zip"
         },
         "Handler": "datadog_lambda.handler.handler",
         "Runtime": "python2.7",
@@ -207,7 +207,7 @@
         "Tags": [
           {
             "Key": "dd_sls_plugin",
-            "Value": "v2.18.0"
+            "Value": "v2.20.0"
           },
           {
             "Key": "service",
@@ -236,8 +236,8 @@
           ]
         },
         "Layers": [
-          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Python27:29",
-          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Extension:5"
+          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Python27:33",
+          "arn:aws:lambda:sa-east-1:464622532012:layer:Datadog-Extension:6"
         ]
       },
       "DependsOn": [
@@ -251,7 +251,7 @@
           "S3Bucket": {
             "Ref": "ServerlessDeploymentBucket"
           },
-          "S3Key": "serverless/dd-sls-plugin-integration-test/dev/1618443590964-2021-04-14T23:39:50.964Z/dd-sls-plugin-integration-test.zip"
+          "S3Key": "serverless/dd-sls-plugin-integration-test/dev/1618590312126-2021-04-16T16:25:12.126Z/dd-sls-plugin-integration-test.zip"
         },
         "Handler": "datadog_lambda.handler.handler",
         "Runtime": "python3.6",
@@ -261,7 +261,7 @@
         "Tags": [
           {
             "Key": "dd_sls_plugin",
-            "Value": "v2.18.0"
+            "Value": "v2.20.0"
           },
           {
             "Key": "service",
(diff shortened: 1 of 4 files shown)


nhinsch:
Hm, this doesn't seem right to me. If the snapshot tests fail that means one of two things:
- There is an actual problem with the plugin and we shouldn't release
- The tests are not reliable -- i.e. they are failing even though nothing is wrong with the plugin
It seems to me that we should fix the underlying problem. If we automatically update the snapshots before we release, that is effectively the same as not running the tests before we release, because the tests will always pass and create new snapshots, even if there was a problem with the plugin. Then we would also have a new problem: we would have saved incorrect snapshots, so now the tests will pass with incorrect behavior and fail with correct behavior.

nhinsch:
It looks like the snapshot tests might be failing because of some version numbers/unique identifiers that are different each time we run the tests. To solve this issue in the other libraries we have used `perl` and/or `sed` to find and replace the unstable identifiers with placeholders like `XXX` that are the same for each test run. It looks like that approach might be needed here to make the tests pass consistently.

zARODz11z:
> Hm, this doesn't seem right to me. If the snapshot tests fail that means one of two things:
> 
> * There is an actual problem with the plugin and we shouldn't release
> * The tests are not reliable -- i.e. they are failing even though nothing is wrong with the plugin
> 
> It seems to me that we should fix the underlying problem. If we automatically update the snapshots before we release, that is effectively the same as not running the tests before we release, because the tests will always pass and create new snapshots, even if there was a problem with the plugin. Then we would also have a new problem: we would have saved incorrect snapshots, so now the tests will pass with incorrect behavior and fail with correct behavior.

As it stands right now the integration tests will be run on any push to any branch however it is not a part of the release process until the release script pushes to the master branch and then the integration tests are run. Currently the integration tests account the layer version numbers as part of the assertion so any time the layers are updated or the version of the plugin is incremented then the snapshots are thrown off.

zARODz11z:
> It looks like the snapshot tests might be failing because of some version numbers/unique identifiers that are different each time we run the tests. To solve this issue in the other libraries we have used `perl` and/or `sed` to find and replace the unstable identifiers with placeholders like `XXX` that are the same for each test run. It looks like that approach might be needed here to make the tests pass consistently.

Yeah so some of the identifiers are unstable during each run of the integration tests but my way around it was to exclude them from the diff via a regex [here](https://github.com/DataDog/serverless-plugin-datadog/blob/04b5ffc4d9670aa60a37e41b8f97526f32c10ffb/scripts/run_integration_tests.sh#L49).

However if you look at the latest commits to the SLS plugin the integration tests outputs why it failed:
* [v2.19.0](https://github.com/DataDog/serverless-plugin-datadog/runs/2356950462): This failed because of the differing SLS plugin version from `v2.18.0 -> v2.19.0`
* [Update layers json](https://github.com/DataDog/serverless-plugin-datadog/runs/2357035317): This failed because of the differing layer versions:
   * Python 27 (v29 to v33)
   * Python 36 (v29 to v33)
   * Python 37 (v29 to v33)
   * Python 38 (v29 to v33)
   * Node10-x (v49 to v52)
   * Node12-x (v49 to v52)
   * Node14-x (v49 to v52)
   * On top of these differences the SLS plugin version was still off from `v2.18.0 -> v2.19.0` .
*  [v2.20.0](https://github.com/DataDog/serverless-plugin-datadog/runs/2357050659): This failed because of the differing SLS plugin version from `v2.18.0 -> v2.20.0`. On top of that the layer versions. were not updated.

nhinsch:
In the other integration tests we do replace version numbers and whatnot with placeholder values (`vX.X` or something like that) and we haven't had any problems. The issue I have with updating the snapshots AFTER releasing would be that technically the released version of the plugin would fail the integration tests -- it would have the new version numbers but the old snapshots.

zARODz11z:
Closing in favor of #126

PR #127, opened 2021-04-22
[Serverless Plugin] Automate forwarder subscriptions for API Gateway log groups

jcstorms1:
### What does this PR do?
Adds the ability for the Serverless plugin to subscribe the forwarder to API Gateway log groups. Previously this was only automated for lambda based log groups. 

- Adds three new environment variables for automatically subscribing to Api Gateway, HTTP API and Websocket log groups
- Adds an API Gateway execution log group and subscription via the plugin
- Allows extension to be used with Forwarder so users can subscribe their API Gateway, Http-Api and Websocket log groups
- Adds unit tests
- Adds additional integration tests
- Updated readme to account for new environment variables

### Motivation

https://datadoghq.atlassian.net/browse/SLS-636

### Testing Guidelines

Unit and integration tests
Deployed to sandbox account

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


scripts/run_integration_tests.sh
```diff
@@ -21,7 +21,7 @@ scripts_dir=$(dirname $script_path)
 repo_dir=$(dirname $scripts_dir)
 root_dir=$(pwd)
 if [[ "$root_dir" =~ .*"serverless-plugin-datadog/scripts".* ]]; then
-    echo "Make sure to run this script from the root `serverless-plugin-datadog` directory, aborting"
+    echo "Make sure to run this script from the root $(serverless-plugin-datadog) directory, aborting"
```
nhinsch, line 24: What's this change do?

integration_tests/correct_forwarder_snapshot.json
```diff
@@ -700,6 +707,345 @@
...
+              [
+                "arn:",
+                {
+                  "Ref": "AWS::Partition"
+                },
+                ":apigateway:",
+                {
+                  "Ref": "AWS::Region"
+                },
+                ":lambda:path/2015-03-31/functions/",
```
zARODz11z, line 778: It would be kinda hard to test if this is changes without waiting a day haha but I feel that this date may also need to be scrubbed

README.md
```diff
@@ -22,24 +22,24 @@ To quickly get started, follow the installation instructions for [Python][1] or
...
+| `enableDDTracing`      | Enable Datadog tracing on the Lambda function. Defaults to `true`.                                                                                                                                                                                                                                                                                                                                                  |
+| `enableApiGatewayLogs` | Enable automatic subscription of the Datadog Forward to API Gateway log groups. Defaults to `true`.                                                                                                                                                                                                                                                                                                                 |
```
nhinsch, line 36: Datadog Forward -> Datadog Forwarder

Also, I'm not sure that `enableApiGatewayLogs` is the best name for this setting. My first thought when I hear it is that this would enable/disable the logs. However, it sounds like the real behavior of this setting is to enable/disable **subscribing the Forwarder** to the API Gateway logs. Maybe it should be called something like `subscribeForwarderToApiGatewayLogs`?

Finally, it seems like this will only work if a `forwarderArn` is provided. We should say that and also probably add some kind of check that will raise an error if you try to enable subscribing to API Gateway logs without providing a forwarder ARN.

integration_tests/correct_forwarder_snapshot.json
```diff
@@ -700,6 +707,345 @@
...
```
zARODz11z, line 821: same as this [comment](https://github.com/DataDog/serverless-plugin-datadog/pull/127/files?file-filters%5B%5D=.json&file-filters%5B%5D=.sh#r618525768)

README.md
```diff
@@ -22,24 +22,24 @@ To quickly get started, follow the installation instructions for [Python][1] or
...
+| `forwarderArn`         | Setting this parameter subscribes the given Datadog forwarder to the Lambda functions’ CloudWatch log groups. Required when `enableDDTracing` is set to `true` unless the subscription is otherwise applied. For example, if a Datadog Forwarder subscription is applied via Datadog's AWS Integration, then `forwarderArn` is not required. Do not set `addExtension` to `true` when using `forwarderArn`.         |
```
nhinsch, line 37: Moving forward, we are going to want customers using the Lambda Extension to send Lambda Logs to Datadog. But if they want API Gateway logs they would need the Forwarder as well for that. So, we would want to enable subscriptions for the Forwarder to the API Gateway logs but *not* use the Forwarder for the Lambda logs. I think there are probably some code/documentation updates needed to cover this situation.

README.md
```diff
@@ -22,24 +22,24 @@ To quickly get started, follow the installation instructions for [Python][1] or
...
+| `exclude`              | When set, this plugin will ignore all specified functions. Use this parameter if you have any functions that should not include Datadog functionality. Defaults to `[]`.                                                                                                                                                                                                                                            |
+| `enabled`              | When set to false, the DataDog plugin will stay inactive. Defaults to `true`. You can control this option using an environment variable, e.g. `enabled: ${strToBool(${env:DD_PLUGIN_ENABLED, true})}`, to activate/deactivate the plugin during deployment. Alernatively, you can also use the value passed in through `--stage` to control this option, [see example.](#disable-plugin-for-particular-environment) |
```
nhinsch, line 42: Nit: I know it's not new, but while we're here we should correct DataDog -> Datadog

integration_tests/correct_forwarder_snapshot.json
zARODz11z: actually thats 2015, for some reason I thought it was todays date. Nvm this is probably fine.

nhinsch: I don't think this will change. 3/31/2015 is not today's date, and according to some Googling it appears that this date represents the version of the Lambda API. So it won't be expected to change/shouldn't be scrubbed.

nhinsch: Good thought though!

README.md
nhinsch: For example, right now you should not set both `forwarderArn` and `addExtension`. But now you might actually want to do that.

scripts/run_integration_tests.sh
jcstorms1: I'm not sure it does anything. Seems the linter changed it.

README.md
jcstorms1: @DarcyRaynerDD is this something I should implement now?

README.md
jcstorms1: Fixed

README.md
```diff
@@ -53,10 +55,11 @@ custom:
     logLevel: "info"
     enableXrayTracing: false
     enableDDTracing: true
+    enableAPIGatewayLogs: true
```
zARODz11z, line 58: It seems that this is the only place where the `enableAPIGatewayLogs` parameter is. It seems like the parameters you added were [these](https://github.com/DataDog/serverless-plugin-datadog/pull/127/files?file-filters%5B%5D=.md&file-filters%5B%5D=.sh&file-filters%5B%5D=.ts&file-filters%5B%5D=.yml#diff-00ee6a506727e42b3fb03361b8216c805ede24dc3985d237de852a73bda59f30R53) ones (`subscribeToApiGatewayLogs`, `subscribeToHttpApiLogs`, and `subscribeToWebsocketLogs`). 

Basically I am wondering is `enableAPIGatewayLogs` a dead parameter? I dont see any addition to the plugin code that supports this parameter. 

Maybe these 3 (`subscribeToApiGatewayLogs`, `subscribeToHttpApiLogs`, and `subscribeToWebsocketLogs`) should be documented in the readme?

integration_tests/correct_forwarder_snapshot.json
```diff
@@ -700,54 +732,917 @@
...
+            "Fn::Sub": "arn:${AWS::Partition}:logs:${AWS::Region}:${AWS::AccountId}:log-group:${WebsocketsLogGroup}"
+          },
+          "Format": "$context.identity.sourceIp $context.identity.caller $context.identity.user [$context.requestTime] \"$context.eventType $context.routeKey $context.connectionId\" $context.requestId"
+        },
+        "DefaultRouteSettings": {
+          "DataTraceEnabled": true,
+          "LoggingLevel": "INFO"
+        },
+        "DeploymentId": {
+          "Ref": "WebsocketsDeploymentMyF6XDJP4uiCx2izd6DzYCSpXtJ6qKoX0mjnNdi2fc"
```
zARODz11z, line 1222: This looks like a randomly generated ID (there seem to be others like it below), it might be worth while to re-run the integration tests locally  `./scripts/run_integration_tests.sh` and make sure that ID is stable so it wont throw off the integration tests later on.

scripts/run_integration_tests.sh
zARODz11z: Yeah im not sure why the linter would change an echo string but it seems like to avoid confusion the backticks `` were fine here

src/index.ts
```diff
@@ -127,6 +127,7 @@ module.exports = class ServerlessPlugin {
           aws,
           datadogForwarderArn,
           config.integrationTesting,
+          config.subscribeToApiGatewayLogs,
```
zARODz11z, line 130: Should you grab the other new  parameters also?
  `subscribeToHttpApiLogs`
  `subscribeToWebsocketLogs`
I think you implemented logic that defaults these parameters to true however does the logic support if the customer wants to opt out. Ie. if they explicitly say this in there serverless.yml:
```
subscribeToApiGatewayLogs: false
subscribeToHttpApiLogs: false
subscribeToWebsocketLogs: false
```

zARODz11z:
I merged this [PR](https://github.com/DataDog/serverless-plugin-datadog/pull/128) to keep serverless versions in sync between local development and whats used on the github workflow. It introduced some merge conflicts but they are resolved now :)

src/env.ts
```diff
@@ -55,6 +60,9 @@ const logLevelEnvVar = "DD_LOG_LEVEL";
 const logForwardingEnvVar = "DD_FLUSH_TO_LOG";
 const ddTracingEnabledEnvVar = "DD_TRACE_ENABLED";
 const logInjectionEnvVar = "DD_LOGS_INJECTION";
+const subToApiGatewayLogsEnvVar = "DD_SUBSCRIBED_TO_API_GATEWAY_LOGS";
```
nhinsch, line 63: For these environment variables, I would use `SUBSCRIBE` instead of `SUBSCRIBED`.

"Subscribe" is an imperative verb - we are telling the plugin to subscribe.

"Subscribed" is an adjective, saying that we are, in fact, subscribed. The problem is that we might not actually be subscribed yet. We want to be subscribed, but we don't know if we actually are. So it's a bit misleading to use that here.

src/forwarder.ts
```diff
@@ -60,6 +60,7 @@ export async function addCloudWatchForwarderSubscriptions(
   aws: Aws,
   functionArn: CloudFormationObjectArn | string,
   integrationTesting: boolean | undefined,
+  subscribeToApiGatewayLogs: boolean,
```
nhinsch, line 63: I think this function is getting a bit long/complicated. It takes five arguments and is almost 100 lines long.

A lot of that length/complexity was already there before this PR, but there's no doubt that this PR is adding more length/complexity. Therefore I think now would be a good time to do a refactor here.

First of all, five arguments is a lot -- a good guideline is to really try to avoid more than three arguments if possible. When I look at a call to this function: `await addCloudWatchForwarderSubscriptions(service as Service, aws, "my-func", false, true)`, it's pretty complicated, and it's not obvious what all of the arguments do, particularly the two booleans at the end.

One possible suggestion is switching the `integrationTesting` argument to an environment variable. That way we don't have to pass it around in the code.

I think there are also opportunities to move code out of this function and into smaller helper functions. I bet there are a lot of ways we could refactor this function to break it down into smaller pieces.

Another thought -- maybe we could have two separate functions -- one for the regular function subscriptions and one for the API Gateway-related ones?

I don't know if it's necessary to spend a huge amount of time totally re-doing everything here and implementing all of these suggestions, but a good rule of thumb would be leaving the code simpler than you found it. Otherwise things are just bound to get more and more complex over time.

jcstorms1: I just saw this but had already thought the same thing. I started to break down the function a bit and you can see that in the most recent commit. Let me know if you think it helps.

zARODz11z:
The reason why that last 10.15 test job is not running is because of #130. It should stop attempting to run 10.15 in the future. I think it just did this because this PR was open before I merged #130. 
So no worries `test (10.15)` is replaced with `test (10.19)`

src/env.ts
```diff
@@ -55,6 +60,9 @@ const logLevelEnvVar = "DD_LOG_LEVEL";
 const logForwardingEnvVar = "DD_FLUSH_TO_LOG";
 const ddTracingEnabledEnvVar = "DD_TRACE_ENABLED";
 const logInjectionEnvVar = "DD_LOGS_INJECTION";
+const subToApiGatewayLogsEnvVar = "DD_SUBSCRIBE_TO_API_GATEWAY_LOGS";
```
DarcyRaynerDD, line 63: I'm not sure DD_SUBSCRIBE_TO_* env vars are useful to the lambda. It doesn't change the behavior of our layers or our extension. Maybe we should remove them?

src/forwarder.spec.ts
```diff
@@ -178,6 +265,153 @@ describe("addCloudWatchForwarderSubscriptions", () => {
...
+              "Ref": "WebsocketGroup",
+            },
+          },
+          "Type": "AWS::Logs::SubscriptionFilter",
+        },
+      }
+    `);
+  });
+
+  it("it does not add subscriptions for log groups that have their subscriptions diabled", async () => {
```
DarcyRaynerDD, line 288:
```suggestion
  it("does not add subscriptions for log groups that have their subscriptions diabled", async () => {
```

src/forwarder.ts
```diff
@@ -140,3 +173,51 @@ export async function describeSubscriptionFilters(aws: Aws, logGroupName: string
     return [];
   }
 }
+
+// Helper functions to validate we have a particular log group and if we should subscribe to it
+function validateApiGatweaySubscription(resource: any, subscribe: boolean) {
```
DarcyRaynerDD, line 178: typo: Needs to be renamed to `validateApiGatewaySubscription`

src/forwarder.ts
```diff
@@ -68,13 +75,35 @@ export async function addCloudWatchForwarderSubscriptions(
...
+        validateHttpApiSubscription(resource, forwarderConfigs.SubToHttpApiLogGroup) ||
+        validateWebsocketSubscription(resource, forwarderConfigs.SubToWebsocketLogGroup)
+      )
+    ) {
+      continue;
+    }
+    // if the extension is disabled, we should subscribe to lambda log groups
+    if (
+      !(
+        resource.Properties.LogGroupName.startsWith("/aws/lambda/") ||
```
DarcyRaynerDD, line 101: When you find yourself writing complex branching conditions with lots of continues and breaks that boil down to a single decision, you can often simplify it to use a function with a bunch of return statements. Eg:
```typescript
function shouldSubscribe(resouce: any, forwarderConfigs: ForwarderConfigs): boolean {
   if (validateXXXXSubscription(resource, forwarderConfigs.SubToApiGatewayLogGroup) ){
      return true;
   }
   //... Other api gateway log groups types here etc etc
   if (!forwarderConfigs.AddExtension && resource.Properties.LogGroupName.startsWith("/aws/lambda/")) {
      return true;
   }
   return false;
}
```

DarcyRaynerDD (changes requested):

src/forwarder.ts
```diff
@@ -140,3 +151,80 @@ export async function describeSubscriptionFilters(aws: Aws, logGroupName: string
...
+  return resource.Properties.LogGroupName.startsWith("/aws/api-gateway/") && subscribe;
+}
+function validateHttpApiSubscription(resource: any, subscribe: boolean) {
+  return resource.Properties.LogGroupName.startsWith("/aws/http-api/") && subscribe;
+}
+function validateWebsocketSubscription(resource: any, subscribe: boolean) {
+  return resource.Properties.LogGroupName.startsWith("/aws/websocket/") && subscribe;
+}
+
+function shouldNotSubscribe(resource: any, forwarderConfigs: ForwarderConfigs) {
```
DarcyRaynerDD, line 166: nit: By convention, we usually use positives for these kinds of functions, eg 'shouldSubscribe'. This is so we avoid using double negatives in our logic, eg 
```typescript
!shouldNotSubscribe() // <-- this is a double negative
shouldNotSubscribe()
// vs
shouldSubscribe()
!shouldSubscribe()
```

DarcyRaynerDD (approved):
Left a minor nit

PR #131, opened 2021-05-06
Add syntax to enable serverless monitors from plugin

hjiang27:
### What does this PR do?

Adds syntax that allows users to enable recommended serverless and custom monitors via the plugin. 
### Motivation

Allows customers to easily enable monitors for their functions while deploying them through the Serverless Framework. See ticket [here](https://datadoghq.atlassian.net/browse/SLS-989). 

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [X] This PR impacts documentation, and it has been updated (or a ticket has been logged)


src/env.ts
```diff
@@ -17,6 +17,10 @@ export interface Configuration {
   apiKey?: string;
   // Datadog API Key encrypted using KMS, only necessary when using metrics without log forwarding
   apiKMSKey?: string;
+  // Datadog API Key used for enabling monitor configuration through plugin
+  monitorsApiKey?: string;
+  // Datadog App Key used for enabling monitor configuratio throguh plugin
```
nhinsch, line 22: configuratio throguh -> configuration through

src/env.ts
```diff
@@ -17,6 +17,10 @@ export interface Configuration {
   apiKey?: string;
   // Datadog API Key encrypted using KMS, only necessary when using metrics without log forwarding
   apiKMSKey?: string;
+  // Datadog API Key used for enabling monitor configuration through plugin
+  monitorsApiKey?: string;
```
nhinsch, line 21: Why is this key different from the normal `apiKey`?

README.md
```diff
@@ -136,6 +136,84 @@ custom:
...
+Define the appropriate serverless monitor ID without specifying any parameter values
+
+```
+custom:
+ datadog:
+   addLayers: true
+   monitorsApiKey: ${file(./config.json):monitorsApiKey}
+   monitorsAppKey: ${file(./config.json):monitorsAppKey}
+   monitors:
+     - high_error_rate:
```
nhinsch, line 167: Is there supposed to be a `:` at the end of this line?

README.md
```diff
@@ -136,6 +136,84 @@ custom:
...
+        options: {
+          thresholds: {
+            warning: 0.05
+          }
+        }
+```
+
+##### To delete a monitor 
+Removing the serverless monitor ID and its parameters will delete the monitor. 
+```
```
nhinsch, line 189: Nit: I'm not sure I would necessarily include an example to show deletion

README.md
```diff
@@ -136,6 +136,84 @@ custom:
   datadog:
     enabled: ${self:custom.staged.dd_enabled, true}
 ```
+### Serverless Monitors 
+
+Serverless has seven recommended monitors with default values pre-configured.
```
nhinsch, line 141: I would reword this... not sure it really makes sense to say that "Serverless" has recommended monitors.

src/monitors.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+  return monitorParams;
+}
+
+/**
+ * Checks to see if the given monitor is a serverless recommended monitor
+ * @param serverlessMonitorId - Unique ID string defined for each serverless monitor
+ * @returns true if a given monitor is a serverless recommended monitor
+ */
+function checkIfServerlessMonitor(serverlessMonitorId: string) {
```
nhinsch, line 66: I would rename this to `checkIfRecommendedMonitor` -- it seems to me that all monitors (custom or recommended) are equally serverless, and what distinguishes these are not that they are serverless but that they are recommended.

src/monitors.ts
```diff
@@ -0,0 +1,180 @@
...
+  return Object.keys(SERVERLESS_MONITORS).includes(serverlessMonitorId);
+}
+
+/**
+ * Checks to see if the monitor already exists
+ * @param serverlessMonitorId - Unique ID string defined for each serverless monitor
```
nhinsch, line 72: Does this argument need to be called `serverlessMonitorId`? What makes a monitor "serverless" as opposed to non-serverless here?

nhinsch: I think this comment applies throughout the PR -- I would look into standardizing the usage of "serverless monitor". To me it looks like in some cases this term basically means "recommended monitor" and in other cases it just seems to mean "monitor" (recommended or custom) -- let's make sure the terminology is consistent throughout.

src/serverless_monitors.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const HIGH_ERROR_RATE: ServerlessMonitor = {
+  name: "High Error Rate on {{functionname.name}} in {{region.name}} for {{aws_account.name}}",
+  query: (cloudFormationStackId: string) => {
+    return `avg(last_15m):sum:aws.lambda.errors{aws_cloudformation_stack-id:${cloudFormationStackId}} by {functionname,region,aws_account}.as_count() / sum:aws.lambda.invocations{aws_cloudformation_stack-id:${cloudFormationStackId}} by {functionname,region,aws_account}.as_count() >= 0.1`;
+  },
+  message:
+    "More than 10% of the function’s invocations were errors in the selected time range. {{#is_alert}} Resolution: Examine the function’s logs, check for recent code or configuration changes with [Deployment Tracking](https://docs.datadoghq.com/serverless/deployment_tracking), or look for failures across microservices with [distributed tracing](https://docs.datadoghq.com/serverless/distributed_tracing).{{/is_alert}}",
+};
+export const TIMEOUT: ServerlessMonitor = {
```
nhinsch, line 16: Nit: I would add a blank line between each of these

src/serverless_monitors.ts
```diff
@@ -0,0 +1,73 @@
...
+  name: "Timeout on {{functionname.name}} in {{region.name}} for {{aws_account.name}}",
+  query: (cloudFormationStackId: string) => {
+    return `avg(last_15m):sum:aws.lambda.duration.maximum{aws_cloudformation_stack-id:${cloudFormationStackId}} by {aws_account,functionname,region}.as_count() / (sum:aws.lambda.timeout{aws_cloudformation_stack-id:${cloudFormationStackId}} by {aws_account,functionname,region}.as_count() * 1000) >= 1`;
+  },
+  message:
+    "At least one invocation in the selected time range timed out. This occurs when your function runs for longer than the configured timeout or the global Lambda timeout. Resolution: [Distributed tracing](https://docs.datadoghq.com/serverless/distributed_tracing) can help you pinpoint slow requests to APIs and other microservices. You can also consider increasing the timeout of your function. Note that this could affect your AWS bill.",
+};
+export const OUT_OF_MEMORY: ServerlessMonitor = {
```
nhinsch, line 24: In the README it says you can customize recommended monitors with a custom `threshold` but these values look like they're hard-coded and I don't see any code or tests for changing the threshold for these monitors. Am I missing something here?

src/monitors.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+export async function setMonitors(
+  monitors: Monitor[],
+  monitorsApiKey: string,
+  monitorsAppKey: string,
+  cloudFormationStackId: string,
+  service: string,
+  env: string,
+) {
+  const serverlessMonitorIdToMonitorId = await getExistingMonitors(
```
nhinsch, line 131: I'm a little bit confused about what this constant means - `serverlessMonitorIdToMonitorId` to me sounds like it would be a function that converts one type of ID to another, but that doesn't seem to be what this is.

src/monitors.ts
```diff
@@ -0,0 +1,180 @@
...
+    cloudFormationStackId,
+    monitorsApiKey,
+    monitorsAppKey,
+  );
+  const succesfullyUpdatedMonitors: string[] = [];
```
nhinsch, line 136: succesfullyUpdatedMonitors -> successfullyUpdatedMonitors

nhinsch:
For the future, I would recommend splitting up a feature like this into multiple pull requests 🙂 1600 lines is a lot of code! 😂  Adding the feature more iteratively can make it easier to review the code and get feedback faster.

src/serverless_monitors.ts
nhinsch: If this feature is indeed not implemented yet, I would suggest taking it out of the README and implementing it later rather than implementing it in this PR -- like I mentioned in another comment, I think there is already a lot going on in this PR!

src/monitors.ts
```diff
@@ -0,0 +1,180 @@
+import { SERVERLESS_MONITORS } from "./serverless_monitors";
+import { updateMonitor, createMonitor, deleteMonitor, getExistingMonitors } from "./monitor-api-requests";
+
+export interface MonitorParams {
+  [key: string]: any;
```
nhinsch, line 5: Can this interface really contain any key/value pair, or are there more specific rules for what is and is not a valid `MonitorParams`?

README.md
hjiang27: Yup, I believe that's part of the serverless.yml syntax so I'm not sure if I can do anything about it, although I do agree it looks a little off.

src/env.ts
hjiang27: I believe Tian and SF said that it would be better to keep the keys separate for security reasons

src/monitors.ts
hjiang27: The syntax is designed to be fairly flexible so that customers can choose which parameters to define and which ones they choose to leave out, so it's difficult to establish specific rules for what is valid and not. Ultimately, if the params are invalid, the API request will fail and it'll log an invalid syntax error message so I thought it would be best to just leave it as any general key/value pair

src/serverless_monitors.ts
hjiang27: That's a good point. Every other param except for query (where the critical threshold is defined) can be overridden if the customer has already defined the parameter in their serverless.yml, but I think it would make more sense if we gave customers the ability to override query so that they can tweak parts of it (including the critical threshold) even if it's hard-coded. If that's the case, it should be a pretty straightforward and small change in the code (I'm thinking a few lines). Do you think it would still be worth looking into opening a separate PR?

README.md
```diff
@@ -136,6 +136,76 @@ custom:
   datadog:
     enabled: ${self:custom.staged.dd_enabled, true}
 ```
+### Serverless Monitors 
+
+There are seven recommended monitors with default values pre-configured. 
+
+| Monitor              | Metrics                                                                                   | Default Threshold | Serverless Monitor ID |
```
nhinsch, line 143: Since the threshold cannot be customized, I would call it the "threshold" rather than the "default threshold"

README.md
```diff
@@ -136,6 +136,76 @@ custom:
...
+| Increased Cost       | `aws.lambda.enhanced.estimated_cost`                                                      | &#8593;20%        | `increased_cost`      |
+ 
+#### To Enable and Configure a Recommended Serverless Monitor 
+
+To create a serverless monitor, you must use its respective serverless monitor ID in addition to passing in the API key and Application key in the `serverless.yml` file. If you’d like to further configure the specific parameter values for a recommended monitor, you can directly define parameters below the serverless monitor ID. Parameters not specified under a recommended monitor will use the default recommended value. At this time, the query parameters for recommended monitors cannot be overriden or customized. As such, the critical threshold value and query for recommended monitors will default to the recommended monitor. For further documentation on monitor parameters, see the [Datadog Monitors API](https://docs.datadoghq.com/api/latest/monitors/#create-a-monitor). You can directly update a monitor by changing the parameter values. To delete a monitor, remove the monitor from the `serverless.yml` template.
```
nhinsch, line 155: I would clarify that you need to pass in the separate `monitorsApiKey` and `monitorsAppKey` parameters -- it's not enough to just set the regular apiKey parameter as currently written.

nhinsch: Also, you mention that the query parameter cannot be overriden or customized. What happens if the customer tries to override the query parameter by accident -- does this fail silently, crash, or something else?

README.md
```diff
@@ -136,6 +136,76 @@ custom:
...
+
+##### To delete a monitor 
+Removing the serverless monitor ID and its parameters will delete the monitor. 
+
+#### To Enable and Configure a Custom Monitor
+
+To define a custom monitor, you must define a unique serverless monitor ID string in addition to passing in the API key and Application key. The `query` parameter is required but every other parameter is optional. 
+
+
+##### To create a custom monitor
```
nhinsch, line 195: I think you could combine this subheading -- it's the only section under "To Enable and Configure a Custom Monitor"

src/env.ts
nhinsch: I am still curious why this is needed -- maybe @tianchu or @sfirrin could shed light here.

It looks like for the normal `apiKey` we support KMS with the `apiKMSKey` option. But now we seem to require a second, separate API key just to use the monitors feature, and this key MUST be passed as plain text.

src/index.ts
```diff
@@ -150,8 +152,25 @@ module.exports = class ServerlessPlugin {
 
   private async afterDeploy() {
     const config = getConfig(this.serverless.service);
+    const service = this.serverless.service.getServiceName();
+    const env = this.serverless.getProvider("aws").getStage();
+
     if (config.enabled === false) return;
-    return printOutputs(this.serverless);
+    if (config.monitors && config.monitorsApiKey && config.monitorsAppKey) {
```
nhinsch, line 159: I think we should add some code to the parameters validation logic that if monitors are being created, that the `monitorsApiKey` and `monitorsAppKey` must also set correctly, and display appropriate error messages to the user. For example, if I define monitors but don't specify a valid monitorsApiKey, I think right now the plugin would just silently skip creating monitors. I think it would be better for the deploy to fail in that circumstance to avoid confusion.

We might also want to validate the API key, I'm not sure if we do that for the regular API key but if so we probably want to do it for the monitors API key as well?

src/monitor-api-requests.ts
```diff
@@ -0,0 +1,158 @@
...
+  query: string;
+  id: number;
+  name: string;
+  tags: string[];
+}
+
+export async function createMonitor(
+  serverlessMonitorId: string,
+  monitorParams: MonitorParams,
+  monitorsApiKey: string,
```
nhinsch, line 24: Instead of passing the `monitorsApiKey` and `monitorsAppKey` into every API call, I think it might be cleaner if we just pulled these values out of the config object within the function. The API key and app key are going to be the same for all API calls each time we deploy, so I don't think we need to re-specify them for every API call.

That way we can reduce the number of arguments being passed to each function! This will help make them more readable.

src/monitor-api-requests.ts
```diff
@@ -0,0 +1,158 @@
...
+    },
+    body: JSON.stringify(monitorParams),
+  });
+
+  if (response.status === 200) {
+    return true;
+  } else if (response.status === 403) {
+    throw new InvalidAuthenticationError("Could not perform request due to invalid authentication");
+  } else if (response.status === 400) {
+    console.log(`Invalid Syntax Error: Could not perform request due to incorrect syntax for ${serverlessMonitorId}`);
```
nhinsch, line 42: If we fail to create a monitor due to invalid syntax, do we want to make the deploy fail?

README.md
hjiang27: It will get overridden by the pre-defined query param. I've updated the documentation to reflect this. Thanks for pointing it out!

README.md
```diff
@@ -136,6 +136,76 @@ custom:
   datadog:
     enabled: ${self:custom.staged.dd_enabled, true}
 ```
+### Serverless Monitors
```
nhinsch, line 139: I would add the new `monitorsApiKey` and `monitorsAppKey` options to the list of parameters, and make sure that the differences in behavior between `monitorsApiKey` and `apiKey` are clear.

src/index.ts
hjiang27: This might be something that's relevant to the product side of the project so I'll bring it up with the rest of the team.

README.md
```diff
@@ -28,6 +28,8 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `apiKMSKey`            | Datadog API Key encrypted using KMS. Use this parameter in place of `apiKey` when `flushMetricsToLogs` is `false` or `addExtension` is `true`, and you are using KMS encryption.                                                                                                                                                                                                                                    |
+| `monitorsApiKey`       | Datadog API Key. Only needed when using plugin to create monitors for your function and when `monitors` is defined. Separate from  `apiKey` with your function,  `monitorsApiKey` is only used to create monitors through the Datadog Monitors API.                                                                                                                                                                                                            |
```
nhinsch, line 31: I would clarify this a bit more:
- You can use the same API key for `apiKey` and `monitorsApiKey` if you want
- You mentioned "your function" but the user could have multiple functions

README.md
```diff
@@ -28,6 +28,8 @@ To further configure your plugin, use the following custom parameters in your `s
 | `site`                 | Set which Datadog site to send data, this is only used when `flushMetricsToLogs` is `false` or `addExtension` is `true`. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com` and `ddog-gov.com`. The default is `datadoghq.com`.                                                                                                                                                               |
 | `apiKey`               | Datadog API Key, only needed when `flushMetricsToLogs` is `false` or `addExtension` is `true`. For more information about getting a Datadog API key, see the [API key documentation][3].                                                                                                                                                                                                                            |
```
nhinsch, line 29: I would clarify that this API key (the API KMS key option as well) adds the Datadog API key directly to the user's Lambda functions as an environment variable.

nhinsch (approved):
Looks good! 🥇 Just a couple more minor suggestions for the README.

PR #136, opened 2021-05-13
Fix error message on unsuccessful monitor creation

hjiang27:
### What does this PR do?

Fix the error message that logs even when the response is 200 OK. See below. 

Screen Shot 2021-05-13 at 7 08 25 AM

### Motivation

Needed to fix the misleading error message that was logged when requests and monitor creation was successful. 
### Testing Guidelines

Manual testing through deploying test function. 
### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/monitor-api-requests.ts
```diff
@@ -75,10 +75,12 @@ export async function searchMonitors(queryTag: string, monitorsApiKey: string, m
...
-  try {
-    throw new Error(`${response.status} ${response.statusText}`);
-  } catch (err) {
-    console.error(err);
+  if (response.status !== 200) {
+    try {
+      throw new Error(`${response.status} ${response.statusText}`);
+    } catch (err) {
+      console.error(err);
+    }
```
nhinsch, line 83: I don't think you need to throw an error here if you are always immediately going to catch it. I think you could just log the error message.

nhinsch (approved):

PR #145, opened 2021-07-12
Fix: Correct exclude functionality

samchungy:
### What does this PR do?

Environment Variables:
- Stops environment variables from being added to the provider
- Adds environment variables for only functions which are not on the exclude list.

Forwarder Subscriptions:
- Create forwarder subscriptions to log groups for only functions which are not on the exclude list.

Tags:
- Create tags for only functions which are not on the exclude list.

Output:
- Show links for only functions which are not on the exclude list.

### Motivation

I was very confused when datadog was displaying a monitoring link for a function which I had "excluded" considering the docs for exclude say:

`When set, this plugin will ignore all specified functions. Use this parameter if you have any functions that should not include Datadog functionality. Defaults to [].`

Additionally, a log group subscription was created for a function which is not meant to have the functionality causing it to appear in my datadog monitor site.

### Change
- Change behaviour of several sections of code using getAllFunctions to retrieve all the functions and instead make them use the result of `findHandlers` which already filters the functions which we want to act on.
- Combined the two tag adding sections. Did some refactoring.
- Added a check on the Logical ID of Log Group to check if it belongs to a function which is not filtered.

### Testing Guidelines

- Unit Tests
- Integration Tests
- Locally

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


tianchu:
@samchungy yeah, this issue definitely make sense to be fixed. However, I wonder if there is a better place where we could easily short circuit all the work this plugin does for "excluded" functions, instead of adding a `if` statement to individual places -- the same issue may come back when someone adding a new functionality without knowing to check `excluded`. 

@DarcyRaynerDD what do you think? Any quick suggestion?

@jcstorms1 can you help follow up on this issue from here?

samchungy:
Yeah I didn't exactly enjoy placing and passing the config object around but it seems that this may require a re-work of the overall structure. I'm happy to give it a shot though.

samchungy:
It might make more sense to create our own getAllFunctions and wrap the getAllFunctions() from Serverless and make it return only functions which are not excluded. Then, we just gotta make the rest of the code use it.

tianchu:
Thanks for your contribution @samchungy 🙇‍♂️ ! We will have someone help review your PR :)

samchungy:
It's not quite yet done -> I want to import a file from the Serverless package but I'll have to write up a type file for it. The`getLogGroupLogicalId` function, ran a bit outta time last night.

samchungy:
> Thanks for your contribution @samchungy 🙇‍♂️ ! We will have someone help review your PR :)

Alright, I think this is alot better than before. Go bananas 🍌

src/forwarder.ts
```diff
@@ -189,6 +192,15 @@ function shouldSubscribe(resource: any, forwarderConfigs: ForwarderConfigs) {
   ) {
     return false;
   }
+
+  // If the log group does not belong to our list of handlers, we don't want to subscribe to it
+  if (
+    resource.Properties.LogGroupName.startsWith("/aws/lambda/") &&
+    !handlers.some(({ name }) => resource.Properties.LogGroupName === `/aws/lambda/${name}`)
```
samchungy, line 199: Hmmm, this might not work because people can override the log group name. I might switch it back to comparing the Logical ID of the group

src/index.spec.ts
```diff
@@ -449,16 +449,18 @@ describe("ServerlessPlugin", () => {
           provider: {
             compiledCloudFormationTemplate: {
               Resources: {
-                FirstGroup: {
+                FirstLogGroup: {
```
samchungy, line 452: https://www.serverless.com/framework/docs/providers/aws/guide/resources/

samchungy:
I don't think I can assign people, @astuyve, I updated the integration tests.

astuyve (approved):
Code review looks good, no major comments to add.

Notes for other reviewers/future me:
QA'd by checking out locally (via `gh pr checkout 145`, easier than setting new upstream and pulling)
Linked via `yarn link` and `yarn link serverless-plugin-datadog` in sls project dir
examined generated CF and ensured deployments worked

tested:
Using `forwarderArn`:
- excluding one function
    - Verify no subscription created for function
    - Verify subscription created for other functions
    - Verify env vars set for other functions
    - Verify `dd_sls_plugin` tag not set for excluded function (in tags config of lambda console)
- excluding a function that doesn't exist
    - Verify it didn't crash

Using `extension: true`:
- excluding one function
    - Verify no layers present for function
    - Verify layers used for other functions
    - Verify env vars set for other functions
    - Verify `dd_sls_plugin` tag not set for excluded function (in tags config of lambda console)
- excluding a function that doesn't exist
    - Verify it didn't crash

In all cases:
- Output of SLS deploy didn't include links for excluded functions

PR #176, opened 2021-09-14
[Serverless plugin] Separate Access and  Execution logging env variables

jcstorms1:
### What does this PR do?

Separate Execution logging from the API Gateway logging environment variable
Create a new env variable for Execution logging
Moved the Execution log group logic to its own function

### Motivation

API Gateway Execution log groups are not maintained by Cloudformation. If a user removes a stack and redeploys using the Serverless Plugin, the user may encounter a `LogGroup already exists` error. Rather than enabling Execution log groups by default, it is now disabled in order to avoid the error and give users more control over their API Gateway logging. If a user chooses to enable Execution logging, they can delete the log group manually if the error is encountered.

### Testing Guidelines

Integration and unit testing. 
Manual deploy with Serverless framework.

### Types of changes

- [x] Bug fix
- [x] New feature
- [x] Breaking change

### Check all that apply

- [x] This PR contains breaking changes that are documented in the description


README.md
```diff
@@ -39,9 +39,10 @@ To further configure your plugin, use the following custom parameters in your `s
...
+| `subscribeToWebsocketLogs`  | Enable automatic subscription of the Datadog Forwarder to Websocket Acess log groups. Defaults to `true`.                                                                                                                                                                                                                                                                                                            |
+| `subscribeToExecutionLogs`  | Enable automatic subscription of the Datadog Forwarder to Execution log groups for REST and Websocket APIs if subscriptions to their access logs are also enabled. Note: Cloudformation does not maintain these log groups. If the stack is removed the log groups will need to be removed manually or the next deploy may be unsuccessful. Defaults to `false`.                                                     |
```
nhinsch, line 45: I'm a bit confused by how these four options work:
- It seems that for each type of API, you can choose separately whether or not you want to subscribe to access log groups. For HTTP and Websocket APIs, the parameter name matches the type of API (there are `subscribeToHTTPApiLogs` and `subscribeToWebsocketLogs` options). There is another parameter called  `subscribeToApiGatewayLogs` -- however, HTTP APIs and Websocket APIs are just types of API Gateway APIs. So, does this parameter override the other two? Or is it for REST APIs, which is the third type of API that doesn't seem to have its own setting?
- The setting for execution logs says that it it subscribes to execution logs for REST APIs and Websocket APIs. What about HTTP APIs, do they not have execution logs? Also, it says you need to have subscriptions to their access logs enabled as well -- why is that?
- Another thought: do customers want to have control over which exact types of APIs they want to subscribe to? E.g. do customers want logs from their HTTP APIs but NOT their Websocket APIs? Or would it be simpler to just have one option? This is more of a product question.

jcstorms1: - I can change API Gateway to Rest access logs to remove the confusion. 
- I should definitely replace the requirement for access logs to be enabled for execution logging to get a subscription. I was using that validation to confirm REST or Websocket logging was enabled, but it would be better to grab that from the provider section of the serverless config instead.
- Http APIs do not support execution logging currently so it is not included.
-  I think the plugin should have this much flexibility. The API Gateways already require additional setup in the `serverless.yml` at both the provider and function level. My understanding is these log groups can be very noisy and they may only be enabled for debugging, which means more flexibility is better if there is only an issue in one log group. @DarcyRaynerDD do you agree?

jcstorms1: I'm also thinking we should split the Execution log group env var into separate variables to give more control to the user. Having all or nothing for execution log groups seems troublesome since they will generate a lot of logs.

README.md
```diff
@@ -25,31 +25,31 @@ Each version of the plugin is published with a [specific set of versions of the
...
+| `subscribeToAccessLogs`    | Enable automatic subscription of the Datadog Forwarder to API Gateway Access log groups. Requires that logging for the respective APIs are enabled in the provider section of the `serverless.yaml` file. Defaults to `true`. ]                                                                                                                                                                                            |
+| `subscribeToExecutionLogs` | Enable automatic subscription of the Datadog Forwarder to Execution log groups for REST and Websocket APIs. Requires that logging for the respective APIs are enabled in the provider section of the `serverless.yaml` file. Note: Cloudformation does not maintain these log groups. If the stack is removed the log groups will need to be removed manually or the next deploy may be unsuccessful. Defaults to `false`. |
```
nhinsch, line 43: Nit: Cloudformation -> CloudFormation

README.md
```diff
@@ -25,31 +25,31 @@ Each version of the plugin is published with a [specific set of versions of the
...
+
```
nhinsch, line 44: I like these new parameters muuuuuuch better! Easy to understand! ✨

integration_tests/serverless-forwarder.yml
```diff
@@ -9,15 +9,20 @@ provider:
...
   lambdaHashingVersion: 20201221
   logs:
-    restApi: true
-    httpApi: true
-    websocket: true
+    restApi:
+      accessLogging: true
+      executionLogging: false
+    httpApi: false
+    websocket:
```
nhinsch, line 16: I would change this from `websocket` to `websocketApi` to be consistent with the other options.

nhinsch: Oh, this is in Serverless Framework 🤦‍♂️ never mind...
Point still stands, but nothing we can do.

src/env.ts
```diff
@@ -56,10 +56,10 @@ export interface Configuration {
   // When set, this plugin will configure the specified monitors for the function
   monitors?: { [id: string]: { [key: string]: any } }[];
 
-  // API Gateway logging
-  subscribeToApiGatewayLogs: boolean;
-  subscribeToHttpApiLogs: boolean;
-  subscribeToWebsocketLogs: boolean;
+  // API Gateway Acess logging
```
nhinsch, line 59: Acess -> Access

src/env.ts
```diff
@@ -56,10 +56,10 @@ export interface Configuration {
...
+  subscribeToAccessLogs: boolean;
+  // API Gateway Execution loggin - handles rest and websocket. Http not supported as of Sept.21
```
nhinsch, line 61: loggin -> logging

src/forwarder.ts
```diff
@@ -63,6 +79,43 @@ async function validateForwarderArn(aws: Aws, functionArn: CloudFormationObjectA
   }
 }
 
+export async function addExecutionLogGroupsAndSubscriptions(
+  service: Service,
+  aws: Aws,
+  functionArn: CloudFormationObjectArn | string,
+) {
+  const extendedProvider = (service.provider as any)?.logs;
```
nhinsch, line 87: Why is this called `extendedProvider`?

src/index.ts
```diff
@@ -115,16 +115,16 @@ module.exports = class ServerlessPlugin {
...
+      SubToAccessLogGroups: config.subscribeToAccessLogs,
+      SubToExecutionLogGroups: config.subscribeToExecutionLogs,
     };
 
     const defaultRuntime = this.serverless.service.provider.runtime;
     const handlers = findHandlers(this.serverless.service, config.exclude, defaultRuntime);
 
     let datadogForwarderArn;
     datadogForwarderArn = setDatadogForwarder(config);
+    this.serverless.cli.log("Setting Datadog Forwarder");
```
nhinsch, line 127: What is this log line for? It looks like it will be printed in all cases, even if `datadogForwarderArn` isn't actually set to anything.

src/index.spec.ts
```diff
@@ -819,5 +819,85 @@ describe("ServerlessPlugin", () => {
       // The service and env tags will be set with the values given in the provider instead
       expect(function_).toHaveProperty("tags", {});
     });
+
+    it("Does not attempt add execution log groups if execution logging env variable is false", async () => {
```
nhinsch, line 823: Technically the plugin's configuration settings aren't environment variables. Environment variables are variables you set outside of a program in your computer's operating system, that the operating system then gives the program access to. For example, when you run:

```
DD_API_KEY=xxx sls deploy
```

DD_API_KEY is an environment variable, and the Serverless Framework is given access to that variable by the operating system.

Environment variables are in the format of KEY=value, they don't usually support nesting/JSON/YML (although you can shove JSON into the value if you want to).

What we have for configuring our plugin is just a configuration YML file that we read, they aren't provided to us by the operating system. The settings aren't environment variables per se. So I would change the references from "environment variables" to just "settings" or "configuration options".

nhinsch: Also, nice tests :+1:

src/forwarder.ts
```diff
@@ -63,6 +79,43 @@ async function validateForwarderArn(aws: Aws, functionArn: CloudFormationObjectA
...
+    // create log group
+    const logGroupName = await createRestExecutionLogGroupName(aws);
+    const executionLogGroupKey = "RestExecutionLogGroup";
+    const executionLogGroupName = addExecutionLogGroup(logGroupName);
+    resources[executionLogGroupKey] = executionLogGroupName;
+    // add subscription
+    const executionSubscription = subscribeToExecutionLogGroup(functionArn, executionLogGroupKey);
+    const executionSubscriptionKey = "RestExecutionLogGroupSubscription";
+    resources[executionSubscriptionKey] = executionSubscription;
+  }
```
nhinsch, line 104: It looks like `executionLogGroupKey` and `executionSubscriptionKey` are constants -- I would suggest moving those up outside of the function since they stay the same every time you call the function. Also, I generally follow the convention of making constants in TypeScript/JS `ALL_CAPS_SNAKE_CASE`.

Note that when I say constant I don't mean anything that uses `const` -- that just means it can't be reassigned to something else. I mean constant in the sense of some kind of string or number with a special meaning.

src/forwarder.ts
```diff
@@ -63,6 +79,43 @@ async function validateForwarderArn(aws: Aws, functionArn: CloudFormationObjectA
...
+  if (restExecutionLoggingIsEnabled(extendedProvider)) {
+    // create log group
+    const logGroupName = await createRestExecutionLogGroupName(aws);
+    const executionLogGroupKey = "RestExecutionLogGroup";
+    const executionLogGroupName = addExecutionLogGroup(logGroupName);
+    resources[executionLogGroupKey] = executionLogGroupName;
+    // add subscription
+    const executionSubscription = subscribeToExecutionLogGroup(functionArn, executionLogGroupKey);
+    const executionSubscriptionKey = "RestExecutionLogGroupSubscription";
+    resources[executionSubscriptionKey] = executionSubscription;
```
nhinsch, line 103: Maybe I am missing something here -- it looks like the `resources` object here is shared across all of the functions. If we set the key to a single subscription, then run the function again, will we overwrite the subscription from the first time we ran the function?

README.md
```diff
@@ -25,31 +25,31 @@ Each version of the plugin is published with a [specific set of versions of the
...
+| `enableDDLogs`             | Enable Datadog log collection for the Lambda function. Note: This setting has no effect on logs sent via the Datadog Forwarder. Defaults to `true`.                                                                                                                                                                                                                                                                        |
+| `subscribeToAccessLogs`    | Enable automatic subscription of the Datadog Forwarder to API Gateway Access log groups. Requires that logging for the respective APIs are enabled in the provider section of the `serverless.yaml` file. Defaults to `true`. ]                                                                                                                                                                                            |
```
nhinsch, line 42: Since we are removing the old parameters, this will be a breaking change and require updating the version number to 3.0.0. However, someone might still use the old parameters by accident. What happens if they do? Can we display some kind of helpful warning or error message?

src/forwarder.ts
jcstorms1: This is because the Serverless class from the types module doesn't include the logs parameter in the provider section. We needed to create a union and extend it with logs, which has a specific type we created.

src/index.ts
jcstorms1: I don't recall adding this. Will remove.

src/index.spec.ts
jcstorms1: This is actually testing the presence of the env variable in `index.ts` at line 136.

src/forwarder.ts
jcstorms1: I don't think so. It might rewrite it in the `serverless_state.json` but it shouldn't have an effect in Cloudwatch. I don't know that there is a way to avoid this based on how the hooks work.

README.md
jcstorms1: Any suggestion on how to do that? Did you mean in the README or in a log message?

src/forwarder.ts
nhinsch: Hmm, but isn't the "extended provider" here JUST the logs config section?

src/forwarder.ts
nhinsch: OK -- if there is a separate `resources` object created for each time we call this function then it should work. If you've tested this with multiple functions and it correctly subscribes to all of their logs then this is fine :+1:

src/index.spec.ts
nhinsch: Right, but unless I'm missing something `subscribeToExecutionLogs` is not an env variable... happy to discuss this one more on Zoom if you want

src/forwarder.ts
jcstorms1: I tested this a bit more and it looks OK.

nhinsch (approved):
This LGTM, make sure to increment the major version when you release this as it is a breaking change.

Issue #175, opened 2021-09-14
After upgrading from 2.27 -> 2.34 getting run running offline


chanan:
## Expected Behavior
Runs correctly in 2.27 in offline mode. Did not try deploy

## Actual Behavior
```
yarn run v1.22.11
$ serverless offline --noAuth --noPrependStageInUrl
 
 Error ---------------------------------------------------
 
  Error: Cannot find module 'serverless/lib/plugins/aws/lib/naming'
  Require stack:
  - C:\Work\Source\marketron-api\node_modules\serverless-plugin-datadog\dist\src\forwarder.js
  - C:\Work\Source\marketron-api\node_modules\serverless-plugin-datadog\dist\src\index.js
  - C:\Users\cbraunstein\AppData\Local\Yarn\Data\global\node_modules\serverless\lib\classes\PluginManager.js
  - C:\Users\cbraunstein\AppData\Local\Yarn\Data\global\node_modules\serverless\lib\Serverless.js
  - C:\Users\cbraunstein\AppData\Local\Yarn\Data\global\node_modules\serverless\scripts\serverless.js
  - C:\Users\cbraunstein\AppData\Local\Yarn\Data\global\node_modules\serverless\bin\serverless.js
      at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
      at Function.Module._load (node:internal/modules/cjs/loader:778:27)
      at Module.require (node:internal/modules/cjs/loader:1005:19)
      at require (node:internal/modules/cjs/helpers:94:18)
      at Object.<anonymous> (C:\Work\Source\marketron-api\node_modules\serverless-plugin-datadog\dist\src\forwarder.js:13:18)
      at Module._compile (node:internal/modules/cjs/loader:1101:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
      at Module.load (node:internal/modules/cjs/loader:981:32)
      at Function.Module._load (node:internal/modules/cjs/loader:822:12)
      at Module.require (node:internal/modules/cjs/loader:1005:19)
      at require (node:internal/modules/cjs/helpers:94:18)
      at Object.<anonymous> (C:\Work\Source\marketron-api\node_modules\serverless-plugin-datadog\dist\src\index.js:45:21)
      at Module._compile (node:internal/modules/cjs/loader:1101:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
[...]

PR #178, opened 2021-09-15
bugfix: Remove dependency on serverless to fix no module error in certain cases

astuyve:
### What does this PR do?
Removes a dependency introduced in #145 which introduced a regression (#175).
In cases where users relied on a globally installed version of Serverless, and did not have a `NODE_PATH` which included globals, CJS would fail to resolve `getLogGroupLogicalName`, as global modules wouldn't be searched.

We don't really need the method from Serverless anyway, as it is quite simple

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/forwarder.ts
+++ b/src/forwarder.ts
@@ -1,6 +1,6 @@
 import { FunctionInfo } from "layer";
 import Service from "serverless/classes/Service";
-import { getLogGroupLogicalId } from "serverless/lib/plugins/aws/lib/naming";
+// import { getLogGroupLogicalId } from "serverless/lib/plugins/aws/lib/naming";
 import Aws = require("serverless/plugins/aws/provider/awsProvider");
 
 const logGroupKey = "AWS::Logs::LogGroup";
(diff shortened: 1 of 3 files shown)


src/forwarder.ts
```diff
@@ -1,6 +1,6 @@
 import { FunctionInfo } from "layer";
 import Service from "serverless/classes/Service";
-import { getLogGroupLogicalId } from "serverless/lib/plugins/aws/lib/naming";
+// import { getLogGroupLogicalId } from "serverless/lib/plugins/aws/lib/naming";
```
agocs, line 3: probably wanna delete this wholesale

agocs (approved):

src/forwarder.ts
astuyve: Yes, I opened the PR a wee bit fast

samchungy:
You can probably remove the extra `types/` folder I added then. As well as the tsconfig change. `"include": ["src/**/*", "types/**/*.d.ts"]` to `"include": ["src/**/*"]`,

PR #204, opened 2021-11-17
SLS-1740: Support DATADOG_API_KEY and DATADOG_APP_KEY environment variables for Datadog serverless plugin

IvanTopolcic:
### What does this PR do?

This PR refactors the serverless plugin to use environment variables and deprecates the monitor API and App keys.

### Motivation

This PR is required for Source Code Integration, and is a good quality of life change. Most users are already using `${env:DATADOG_API_KEY}` (or equivalent) in their serverless.yml, and this makes it less confusing when we do enable source code integration.

### Testing Guidelines

I've re-deployed a lambda with the serverless plugin testing for the behavior in various conditions (only DATADOG_API_KEY set, multiple keys set, etc..) and it seems to work fine. Additionally, I've tried to cover most cases in the unit tests.
### Additional Notes

This PR refactors the plugin to remote `monitorsApiKey` and `monitorsAppKey`. Instead, it will pull values from the environment, namely `DATADOG_API_KEY` and `DATADOG_APP_KEY`. It will still accept `monitorsApiKey` and `monitorsAppKey` for for backwards compatibility.

When `DATADOG_API_KEY` is specified, the value will be used and attached to the lambda as `DD_API_KEY`, **unless** they've specified `apiKey` in the config.

Additionally, when `DATADOG_API_KEY` is specified, `DD_API_KEY` will **not** be set if the user specifies an `apiKMSKey`, or `apiKeySecretArn` in the configuration. We need this behavior because users will still need to provide `DATADOG_API_KEY` to do things such as source code integration and monitors, however, might still want to use a KMS key or secret ARN on their lambda. In this case, `DATADOG_API_KEY` will be used to deploy source code integration and monitors, but only `DD_KMS_API_KEY` will be attached to their lambda, and `DD_API_KEY` will not be set.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


README.md
```diff
@@ -33,16 +33,13 @@ To further configure your plugin, use the following custom parameters in your `s
...
+| `apiKeySecretArn`           | The ARN of the secret storing the Datadog API key in AWS Secrets Manager. Use this parameter if you're storing your secrets in AWS Secrets Manager, and when `flushMetricsToLogs` is `false` or `addExtension` is `true`. If set, remember to add the `secretsmanager:GetSecretValue` permission to the Lambda execution role.                                                                                                           |
+| `apiKMSKey`                 | Datadog API Key encrypted using KMS. Use this parameter if you're storing your secrets via KMS, and if `flushMetricsToLogs` is `false` or `addExtension` is `true`. Defining `apiKMSKey` will add the Datadog API Key directly to your Lambda functions as an environment variable.                                                                                                                                                      |
```
sarina-dd, line 39:
```suggestion
| `apiKMSKey`                 | Datadog API Key encrypted using KMS. Use this parameter if you're storing your secrets with KMS, and if `flushMetricsToLogs` is `false` or `addExtension` is `true`. Defining `apiKMSKey` will add the Datadog API Key directly to your Lambda functions as an environment variable.                                                                                                                                                      |
```

README.md
```diff
@@ -85,6 +79,8 @@ custom:
 
 **Note**: If you use webpack, Datadog recommends using the prebuilt layers by setting `addLayers` to `true`, which is the default, and add `datadog-lambda-js` and `dd-trace` to the [externals][6] section of your webpack config.
 
+**Note**: Many of the above configuration values require setting `DATADOG_API_KEY` in your environment. If you store these values in AWS Secrets Manager or AWS KMS, please refer to the above options `apiKeySecretArn` and `apiKMSKey` to pull your Datadog API key from the correct location.
```
sarina-dd, line 82:
```suggestion
**Note**: Many of the above configuration values require setting `DATADOG_API_KEY` in your environment. If you store these values in AWS Secrets Manager or AWS KMS, refer to the above options `apiKeySecretArn` and `apiKMSKey` to pull your Datadog API key from the correct location.
```

README.md
```diff
@@ -222,14 +214,12 @@ Removing the serverless monitor ID and its parameters will delete the monitor.
 
 #### To Enable and Configure a Custom Monitor
 
-To define a custom monitor, you must define a unique serverless monitor ID string in addition to passing in the API key and Application key. The `query` parameter is required but every other parameter is optional. Define a unique serverless monitor ID string and specify the necessary parameters below. For further documentation on monitor parameters, see the [Datadog Monitors API](https://docs.datadoghq.com/api/latest/monitors/#create-a-monitor).
+To define a custom monitor, you must define a unique serverless monitor ID string in addition to passing in the API key and Application key via `DATADOG_API_KEY` and `DATADOG_APP_KEY` in your environment. The `query` parameter is required but every other parameter is optional. Define a unique serverless monitor ID string and specify the necessary parameters below. For further documentation on monitor parameters, see the [Datadog Monitors API](https://docs.datadoghq.com/api/latest/monitors/#create-a-monitor).
```
sarina-dd, line 217:
```suggestion
To define a custom monitor, you must define a unique serverless monitor ID string in addition to passing in the API key and Application key, `DATADOG_API_KEY` and `DATADOG_APP_KEY`, in your environment. The `query` parameter is required but every other parameter is optional. Define a unique serverless monitor ID string and specify the necessary parameters below. For further documentation on monitor parameters, see the [Datadog Monitors API](https://docs.datadoghq.com/api/latest/monitors/#create-a-monitor).
```

sarina-dd (approved):
Docs LGTM 👍

README.md
```diff
@@ -33,16 +33,13 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `site`                      | Set which Datadog site to send data, this is only used when `flushMetricsToLogs` is `false` or `addExtension` is `true`. Possible values are `datadoghq.com`, `datadoghq.eu`, `us3.datadoghq.com`, `us5.datadoghq.com`, and `ddog-gov.com`. The default is `datadoghq.com`.                                                                                                                                                              |
-| `apiKey`                    | Datadog API Key, only needed when `flushMetricsToLogs` is `false` or `addExtension` is `true`. Defining `apiKey` will add the Datadog API key directly to your Lambda functions as an environment variable. For more information about getting a Datadog API key, see the [API key documentation][3].                                                                                                                                    |
```
DarcyRaynerDD, line 38: Like we talked about in person, I think it's a bad idea to delete these keys. People often write their serverless.yaml files to reference secrets using different mechanisms, (eg, it's possible to read values from KMS, or as a command line argument), and need to be able to set the apiKey/appKey as a property inside the template, not just as an environment variable. We should warn people not to commit the keys in plaintext if they do use these field.

duncanista:
I think it would be idea to add in the docs an Environment Variables configuration section, just like in [datadog-ci](https://github.com/DataDog/datadog-ci/tree/master/src/commands/lambda#configuration). What do you think about this?

README.md
duncanista: I agree with not deleting the keys from the docs. We could add a warning/deprecated message just like Darcy suggested. Maybe in the note?

src/index.ts
```diff
@@ -258,13 +264,24 @@ function validateConfiguration(config: Configuration) {
...
   }
   if (config.monitors) {
-    if (config.monitorsApiKey === undefined || config.monitorsAppKey === undefined) {
-      throw new Error("When `monitors` is defined, `monitorsApiKey` and `monitorsAppKey` must also be defined");
+    if (
+      (process.env.DATADOG_API_KEY === undefined || process.env.DATADOG_APP_KEY === undefined) &&
+      // Support deprecated monitorsApiKey and monitorsAppKey
+      (config.monitorsApiKey === undefined || config.monitorsAppKey === undefined)
+    ) {
+      throw new Error("When `monitors` is enabled, `DATADOG_API_KEY` and `DATADOG_APP_KEY` must be set.");
```
duncanista, line 284: What about adding ```... and `DATADOG_APP_KEY` environment variables must be set.```

README.md
IvanTopolcic: Yep, on it!

src/index.ts
IvanTopolcic: Good idea, doing this now!

Issue #210, opened 2021-12-02
arm64 extension layer not applied to nodejs runtime


dillon-sellars-voya:
## Expected Behavior

When using `architecture: arm64` and runtime `nodejs14.x` the extension layer should be Datadog-Extension-ARM. It seems that python has a different runtime for x64 vs arm64 but nodejs14 does not. When it comes to the extension layer, the runtime should probably be ignored and always use the architecture of the lambda. 

## Actual Behavior

The extension layer is Datadog-Extension and the lambda fails to load / rune

## Steps to Reproduce the Problem

  1. Choose node runtime with architecture arm64
  1. sls deploy
  1. Check extension is selected

## Specifications

  - Serverless Framework version:  2.63.0
  - Datadog Serverless Plugin version: 3.30
  - Lambda function runtime (Python 3.7, Node 10, etc.):  nodejs14.x

PR #211, opened 2021-12-02
 arm64 extension layer not applied to nodejs runtime fixes#210

dillon-sellars-voya:
- Use arm64 extensions layer when architecture is arm64 - don't check runtime architecture

### What does this PR do?

When applying the extension layer, don't check the runtime architecture - nodejs just has 1 layer for both architectures. 

### Motivation

https://github.com/DataDog/serverless-plugin-datadog/issues/210

### Testing Guidelines

- Observed incorrect behavior in deployments - nodejs14.x with arm64 used x64 datadog extension layer. 
- Applied fix and deployed - x64 layer replaced with arm64 layer
- Re-created scenario in unit tests and passed unit tests

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/layer.ts
+++ b/src/layer.ts
@@ -113,13 +113,12 @@ export function applyExtensionLayer(service: Service, handlers: FunctionInfo[], 
     if (handler.type === RuntimeType.UNSUPPORTED) {
       continue;
     }
-    const { runtime } = handler;
     const architecture =
       (handler.handler as any).architecture ?? (service.provider as any).architecture ?? DEFAULT_ARCHITECTURE;
     let extensionLayerARN: string | undefined;
     let extensionLayerKey: string = "extension";
 
-    if (architecture === ARM64_ARCHITECTURE && runtime && runtime in armRuntimeKeys) {
+    if (architecture === ARM64_ARCHITECTURE) {
       removePreviousLayer(service, handler, regionRuntimes[extensionLayerKey]);
       extensionLayerKey = armRuntimeKeys[extensionLayerKey];
     }
--- a/src/layer.spec.ts
+++ b/src/layer.spec.ts
@@ -468,6 +468,36 @@ describe("applyLambdaLibraryLayers", () => {
     });
   });
 
+  it("adds correct lambda layer given architecture in provider level for node", () => {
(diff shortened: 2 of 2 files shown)


duncanista (approved):
LGTM! 👍🏽

src/layer.ts
```diff
@@ -113,13 +113,12 @@ export function applyExtensionLayer(service: Service, handlers: FunctionInfo[],
...
       continue;
     }
-    const { runtime } = handler;
     const architecture =
       (handler.handler as any).architecture ?? (service.provider as any).architecture ?? DEFAULT_ARCHITECTURE;
     let extensionLayerARN: string | undefined;
     let extensionLayerKey: string = "extension";
 
-    if (architecture === ARM64_ARCHITECTURE && runtime && runtime in armRuntimeKeys) {
+    if (architecture === ARM64_ARCHITECTURE) {
```
agocs, line 121: @duncanista , I have a question about this line. Will `architecture` always be `ARM64_ARCHITECTURE` if `runtime` is in `armRuntimeKeys`? If so, does it make sense to get rid of `armRuntimeKeys` for being redundant? And if not, does it make sense to add `nodejs14.x` to `armRuntimeKeys`?

I think this PR is probably good, I think I'm just a little confused about `armRuntimeKeys` and the nodejs14.x runtime.

duncanista: Not necessarily. Although every runtime in the `armRuntimeKeys` will always have an `ARM64_ARCHITECTURE`. Not every architecture will have a runtime in that map. `armRuntimeKeys` is needed since some runtimes, like `python3.8` and `python3.9`, need its specific ARM library layer. For example, `python3.7` and node runtimes do not need it since their library works right out of the box. Therefore, we don't need to add `nodejs14.x` or any other node environment to `armRuntimeKeys`.

agocs (approved):
LGTM also! Integration tests pass locally.

agocs:
@dillon-sellars-voya Thanks so much for this contribution. We're waiting for a new version of the Lambda extension to be released, and then we'll release this as well.

Issue #223, opened 2022-02-01
Cannot find module 'axios'


gs11:
## Expected Behavior

`sls package` successfully builds a package

## Actual Behavior

Fails with `Cannot find module 'axios'`.
After manually installing it, it then fails with not finding `form-data`

## Steps to Reproduce the Problem

  1. Add `"serverless-plugin-datadog": "^3.5.2",` as a dev dependency to your sls service
  1. Run `npm install`
  1. Run `sls package`

## Specifications

  - Serverless Framework version: 
Framework Core: 2.66.2
Plugin: 5.5.1
SDK: 4.3.0
Components: 3.18.1
  - Datadog Serverless Plugin version: 3.5.2
  - Lambda function runtime (Python 3.7, Node 10, etc.): v16.13.0
  
## Stacktrace
``` Error ---------------------------------------------------

  Error: Cannot find module 'axios'
  Require stack:
  - /redacted-my-service/node_modules/serverless-plugin-datadog/dist/src/helpers/apikey.js
  - /redacted-my-service/node_modules/serverless-plugin-datadog/dist/src/source-code-integration.js
  - /redacted-my-service/node_modules/serverless-plugin-datadog/dist/src/index.js
  - /home/gs/.nvm/versions/node/v16.13.0/lib/node_modules/serverless/lib/classes/PluginManager.js
  - /home/gs/.nvm/versions/node/v16.13.0/lib/node_modules/serverless/lib/Serverless.js
  - /home/gs/.nvm/versions/node/v16.13.0/lib/node_modules/serverless/scripts/serverless.js
  - /home/gs/.nvm/versions/node/v16.13.0/lib/node_modules/serverless/bin/serverless.js
      at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
      at Function.Module._load (node:internal/modules/cjs/loader:778:27)
      at Module.require (node:internal/modules/cjs/loader:1005:19)
      at require (node:internal/modules/cjs/helpers:102:18)
      at Object.<anonymous> (/redacted-my-service/node_modules/serverless-plugin-datadog/dist/src/helpers/apikey.js:16:33)
      at Module._compile (node:internal/modules/cjs/loader:1101:14)
      at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
      at Module.load (node:internal/modules/cjs/loader:981:32)
      at Function.Module._load (node:internal/modules/cjs/loader:822:12)
[...]

PR #224, opened 2022-02-01
Add required dependencies

IvanTopolcic:
Fixes issue #223 by adding missing dependencies, and removing unnecessary dependencies.

I've also changed the imports of files to be relative, so that it can pass `depcheck`. Imports of external dependencies should use `import x from "y"` while imports from local files should use `import x from "./y"`. 

This PR also adds a CI job that uses [depcheck](https://www.npmjs.com/package/depcheck) to determine if we're using dependencies that haven't been added to the package.json file.


src/helpers/utils.ts
```diff
@@ -94,6 +94,7 @@ export const getRequestBuilder = (options: RequestOptions) => {
 
     if (options.headers !== undefined) {
       options.headers.forEach((value, key) => {
+        // @ts-ignore
```
maxday, line 97: why do we need this?

tsconfig.json
```diff
@@ -4,7 +4,7 @@
     // "incremental": true,                   /* Enable incremental compilation */
     "target": "es6" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    "lib": ["es2015"] /* Specify library files to be included in the compilation. */,
+    "lib": ["es2015", "dom"] /* Specify library files to be included in the compilation. */,
```
DarcyRaynerDD, line 7: Why do we need to add dom libraries?

IvanTopolcic: I'm actually not entirely sure here -- when trying to build the project, I was running into an error with Axios: `Cannot find name 'AbortSignal'.`. Importing `dom` fixed this. Is this perhaps a versioning error or something else?

src/helpers/utils.ts
IvanTopolcic: Typescript was complaining about an `any` type here and this seemed like the simplest solution as the code was taken from datadog-ci and I wasn't very familiar with it.

tsconfig.json
IvanTopolcic: Weirdly enough it seems to be working now without it, not sure what that was, but I've gone ahead and removed it for now.

src/helpers/utils.ts
DarcyRaynerDD: It's probably better to add an explicit type. If you're not sure what that is, you can use an any type.

nhinsch:
Per Slack convo: let's add something to the CI pipeline to catch this kind of issue in the future.

src/helpers/utils.ts
IvanTopolcic: Done!

scripts/depcheck.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+DEPCHECK_OUTPUT=`depcheck`
+
+if [[ `echo $DEPCHECK_OUTPUT | grep Missing | wc -l` -gt 0 ]]; then
+    echo "Found some missing dependencies."
+    echo "$DEPCHECK_OUTPUT"
+    exit 1
+    echo "`yarn add` the missing dependencies and re-commit package.json."
```
nhinsch, line 9: I think this line will never print since we `exit 1` right above it

IvanTopolcic: ah dumb mistake, thank you!

scripts/depcheck.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
nhinsch, line 1: Why do we need this shell script? Can we just call depcheck directly?

IvanTopolcic: Depcheck returns a bunch of other things, such as dependencies that are imported but not used, and apparently this is not 100% accurate so I did a little bit of checking for only un-imported dependencies. For instance, it does not think we are using jest.

IvanTopolcic: For instance, right now it's returning this:

```
Unused devDependencies
* @types/jest
```

which I think is safe to ignore because we shouldn't care too much about unused dependencies as they're not going to break anything

nhinsch: Are we in fact not using that devDependency? Can we remove it? I think it would be cleaner to fix any issues depcheck finds rather than adding a wrapper around it. On the other hand, if depcheck is giving inaccurate results that is concerning 🤔

nhinsch: If there's some issue specifically around Jest, it looks like depcheck has an `--ignores` option which might work as well

IvanTopolcic: Hmm interesting, that's a good catch. I did end up removing it and it seems to be passing. I had thought that I tested this and the tests had failed but that might have been an unrelated issue. Apparently there can be false positives, though. I've fixed the issue and tested it, should be a lot cleaner now.

> The logic of a special is not perfect. There might be false alerts. If this happens, please open an issue for us.

IvanTopolcic: Okay yeah, it looks like the tests are failing in CI now but not locally after removing `@types/jest`. Not really sure what to do here.

IvanTopolcic: I'm just going to ignore `@types/jest` for now.

jest.config.js
```diff
@@ -1,16 +0,0 @@
-// For a detailed explanation regarding each configuration property, visit:
```
nhinsch, line 1: Why is this file deleted?

IvanTopolcic: Jest configurations can be specified either in the package.json file, or a jest.config.js file. Since we already have settings for it in the package.json file, it was redundant to have this file so I deleted it.

nhinsch (approved):
Looks great! 🌟

PR #236, opened 2022-03-04
[SLS-1618] Fix environment variable override logic

lizapressman:
### What does this PR do?

DD_TRACE_ENABLED was getting set correctly in env.ts but tracing.ts we were overriding the value set at the handler level with the value of the config so this PR removes that logic

### Motivation

https://datadoghq.atlassian.net/browse/SLS-1618

### Testing Guidelines

Tested locally and deployed testing function and looks to be working as expected

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/env.ts
```diff
@@ -151,6 +152,9 @@ export function setEnvConfiguration(config: Configuration, handlers: FunctionInf
     if (config.enableDDTracing !== undefined && environment[ddTracingEnabledEnvVar] === undefined) {
       environment[ddTracingEnabledEnvVar] = config.enableDDTracing;
     }
+    if (config.enableXrayTracing !== undefined && environment[ddMergeXrayTracesEnvVar] === undefined) {
```
maxday, line 155: Maybe you could extract this logic in a pure function, and add unit tests around it?

lizapressman: Do you mean extract into a function like this?

```
function shouldUpdateEnvironment(configVariable: boolean | undefined, environmentVariable: undefined) {
  return configVariable !== undefined && environmentVariable === undefined;
}
```

I'm totally happy to extract it and write tests but I'm not sure I totally follow why that would be necessary since we already test the logic with the tests we currently have in `env.spec,ts` and this seems like a fairly trivial undefined check.

maxday (approved):

src/tracing.ts
```diff
@@ -27,12 +27,12 @@ export function enableTracing(service: Service, tracingMode: TracingMode, handle
...
   }
-  if (tracingMode === TracingMode.HYBRID || tracingMode === TracingMode.DD_TRACE) {
-    handlers.forEach(({ handler }) => {
-      handler.environment ??= {};
-      const environment = handler.environment as any;
-      environment[ddTraceEnabledEnvVar] = true;
-      environment[ddMergeXrayTracesEnvVar] = tracingMode === TracingMode.HYBRID;
-    });
-  }
+  handlers.forEach(({ handler }) => {
```
astuyve, line 30: Can you help me understand why we removed the conditional check for either tracing mode? Is that somehow handled by the `environment` variable instead?

src/tracing.ts
```diff
@@ -27,12 +27,12 @@ export function enableTracing(service: Service, tracingMode: TracingMode, handle
...
+    handler.environment ??= {};
+    const environment = handler.environment as any;
+    // if tracing is not enabled, merge x-ray cannot be enabled
+    if (environment[ddTraceEnabledEnvVar] === false || environment[ddTraceEnabledEnvVar] === "false") {
```
astuyve, line 34: If a customer had specifically enabled xray merging on a function-level, would they have to set `ddTraceEnabled: true` as well as `ddMergeXrayTraces: true`?

astuyve (commented):
Couple quick questions.

Also a suggestion when working in OSS repos: if you mention the Github Issue in the PR description, the issue will be closed automatically when the PR is merged.

src/tracing.ts
lizapressman: They wouldn't need to specifically set ddTraceEnabled because it will default to true, however, if they have that set as false and had set ddMergeXrayTraces to true (or DD_MERGE_XRAY_TRACES: true at the function-level), DD_MERGE_XRAY_TRACES will get set to false here instead since you can't have DD_MERGE_XRAY_TRACES: true and DD_TRACE_ENABLED: false. 

And they wouldn't need to set ddMergeXrayTraces: true if they've set DD_MERGE_XRAY_TRACES: true at the function level.

src/tracing.ts
lizapressman: The tracingMode was defined by the config (not at the function-level) so checking tracingMode isn't necessary because the config might not match the function-level variables

astuyve: Makes sense!

src/tracing.ts
astuyve: AH gotchya. Sounds good!

astuyve (approved):

PR #192, opened 2021-10-20
Fail on mon creation

ojongerius:
### What does this PR do?

Raise an exception when failing to update monitors

### Motivation

Failing to create monitors is difficult to detect and diagnose at the moment. This makes it awkward to use in a pipeline and hard to troubleshoot.

### Testing Guidelines

Have an invalid monitoring configuration, run `sls deploy --stage foo`, validate that:

* exit status is non zero
* error contains status code

### Types of changes

Bug fix

### Check all that apply

- [x] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
- [ ] This PR's changes are covered by the automated tests
- [ ] This PR collects user input/sensitive content into Datadog

PR #241, opened 2022-03-25
feat: Support failOnError

astuyve:
### What does this PR do?

Optionally allows to fail a serverless deploy command if monitors cannot be set (due to API error, syntax error, etc)

### Motivation

This PR supersedes https://github.com/DataDog/serverless-plugin-datadog/pull/192, which was breaking.

### Testing Guidelines
Set `failOnError` to `true` and put some garbage in a monitor config block. It should throw.

### Additional Notes
Since monitor creation happens after deploy, this is only a way to notify users in CI that something went wrong.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply


README.md
```diff
@@ -58,6 +58,7 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `customHandler`               | When set, the specified handler is set as the handler for all the functions. By default, the handler is set to `/opt/nodejs/node_modules/datadog-lambda-js/handler.handler` if `addLayers` is set to `true`, or `node_modules/datadog-lambda-js/dist/handler.handler` if `addLayers` is set to `false`.                                                                                                                             |
+| `failOnError`                 | When set, this plugin will throw an error if any custom Datadog monitors fail to create or update. This occurs after deploy, but will cause the result of `serverless deploy` to return a nonzero exit code (to fail user CI). Defaults to `false`.                                                                                                                                                                                 |
```
zARODz11z, line 61: In other words when we do fail, we still deploy the resources but the sls deploy command shows failed because it gets a nonzero exit code?

src/index.spec.ts
```diff
@@ -5,9 +5,17 @@
  * This product includes software developed at Datadog (https://www.datadoghq.com/).
  * Copyright 2021 Datadog, Inc.
  */
-
+jest.mock("./monitors.ts", () => {
+  return {
+    setMonitors: async (shouldThrow: Boolean) => {
```
zARODz11z, line 10: I'm a bit confused, how does this mock work? Does this mock return in every test that touches monitors.ts? I guess what I mean is, since in the test you set [failOnError to true](https://github.com/DataDog/serverless-plugin-datadog/pull/241/files#diff-18466f9d870e2151722b884f1cb0df6c1530517bde9468ec2a5abbbb8f655210R738) does that automagically invoke this mock?

README.md
astuyve: That's correct

src/index.spec.ts
astuyve: You're right to be confused. Jest, the test system we're using, rewires how nodejs's `module` system and lets us change methods imported by files.

In this case, we're telling Jest to rewire the `ServerlessPlugin`'s dependency on `monitors.ts`, to instead call our mock. Since I don't want it to error every time (but I also don't want to call the datadog backend when we run these tests), I just pass a boolean if a mocked method should throw or not.

Hence, a mock.

src/index.ts
```diff
@@ -215,6 +215,9 @@ module.exports = class ServerlessPlugin {
       } catch (err) {
         if (err instanceof Error) {
           this.serverless.cli.log(`Error occurred when configuring monitors: ${err.message}`);
+          if (config.failOnError) {
+            throw err;
```
zARODz11z, line 219: Should we log something? Or does the throw on line 219 get caught by line 216?

astuyve: Great question! Since this plugin is for the Serverless framework, we're using `this.serverless.cli.log` to print the message to the command line.

Previously we'd consider this a non fatal error and continue, this PR changes this behavior to make the error fatal (and thus fail CI like github actions, jenkins, gitlab, etc) - only if the configuration option is set.

urseberry (approved):

zARODz11z (approved):

PR #262, opened 2022-05-12
Don't configure tracing for external API Gateway

ifeltsweet:
Without this change deploy fails with the following error:

```
When external API Gateway resource is imported via "provider.apiGateway.restApiId", property "provider.tracing.apiGateway" is ineffective.
```

From the [serverless documentation](https://www.serverless.com/framework/docs/deprecations#aws-api-gateway-schemas):

> When external API Gateway resource is used and imported via provider.apiGateway.restApiId setting, both provider.logs.restApi and provider.tracing.apiGateway are ignored. In v3, an error will be thrown if these options are defined. Indeed, these settings are applicable only if API Gateway resource is provisioned by Serverless Framework.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/tracing.ts
+++ b/src/tracing.ts
@@ -23,7 +23,7 @@ export function enableTracing(service: Service, tracingMode: TracingMode, handle
   const provider = service.provider as any;
   if (tracingMode === TracingMode.XRAY || tracingMode === TracingMode.HYBRID) {
     provider.tracing = {
-      apiGateway: service.provider.apiGateway.restApiId ? null : true,
+      apiGateway: service.provider['apiGateway']?.restApiId ? null : true,
       lambda: true,
     };
   }
--- a/src/tracing.ts
+++ b/src/tracing.ts
@@ -23,7 +23,7 @@ export function enableTracing(service: Service, tracingMode: TracingMode, handle
   const provider = service.provider as any;
   if (tracingMode === TracingMode.XRAY || tracingMode === TracingMode.HYBRID) {
     provider.tracing = {
-      apiGateway: true,
+      apiGateway: service.provider.apiGateway.restApiId ? null : true,
       lambda: true,
     };
   }
--- a/src/tracing.ts
+++ b/src/tracing.ts
@@ -7,6 +7,7 @@
  */
 
 import Service from "serverless/classes/Service";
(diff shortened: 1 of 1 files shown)


ifeltsweet:
I've updated the PR with types fixed. Tests pass now.

Btw here's another report in Serverless repo for this issue https://github.com/serverless/serverless/discussions/10456

src/tracing.ts
```diff
@@ -20,10 +21,12 @@ export enum TracingMode {
 }
 
 export function enableTracing(service: Service, tracingMode: TracingMode, handlers: FunctionInfo[]) {
-  const provider = service.provider as any;
+  const provider = service.provider as Provider;
```
astuyve, line 24: Thanks for fixing the types here!

src/tracing.ts
```diff
@@ -20,10 +21,12 @@ export enum TracingMode {
...
   if (tracingMode === TracingMode.XRAY || tracingMode === TracingMode.HYBRID) {
     provider.tracing = {
-      apiGateway: true,
+      apiGateway: provider.apiGateway?.restApiId
```
astuyve, line 27: Instead of leaving an empty key here, what would you think about changing the logic around?

We could do:
```
provider.tracing = { lambda: true }

if (!provider.apiGateway?.restApiId) {
  provider.tracing.apiGateway = true
}
```

Thoughts?

ifeltsweet: Yes, that would be ideal, but unfortunately, the problem is with the type not allowing to do that.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b6e1b73743d2fdf87d61b0e667e015835aa021f6/types/serverless/plugins/aws/provider/awsProvider.d.ts#L269

I thought it will be a lesser evil to typecast `undefined` to `any` rather than to typecast the whole object or resort to `@ts-ignore`.

I've submitted PR to the types repo https://github.com/DefinitelyTyped/DefinitelyTyped/pull/60489

astuyve: Ha! Okay, thanks for that context. I doubt that the Serverless Framework maintainers will change this behavior to interpret `undefined` as something else, so I think we can accept this patch. Thanks again!

PR #299, opened 2022-09-14
Fix integration test flakiness

sfirrin:
### What does this PR do?

Adds a python script that is used by the integration test script to compare the snapshots instead of using diff

Example output when the snapshot was manually broken:
```
Performing diff of test_extension_snapshot.json against correct_extension_snapshot.json
Values AWS::Lambda::LayerVersion and AWS::Lambda::LayerVersion-BROKEN do not match
Mismatch: Dict values at key Type do not match
Mismatch: Dict values at key ProviderLevelLayerLambdaLayer do not match
Mismatch: Dict values at key Resources do not match
FAILURE: There were differences between the test_extension_snapshot.json and correct_extension_snapshot.json. Review the diff output above.
If you expected the test_extension_snapshot.json to be different generate new snapshots by running this command from a development branch on your local repository: 'UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh'
```

### Motivation

Integration tests were flaky because they were comparing json files with non-deterministic ordering of object keys, so equivalent snapshots were failing the tests due to ordering differences

### Testing Guidelines

Tested by running the integration tests locally

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/scripts/compare_snapshots.py
+++ b/scripts/compare_snapshots.py
@@ -1,0 +1,88 @@
+import json
+import sys
+import logging
+
+log = logging.getLogger(__name__)
+
+
+def compare_without_order(a, b, file_name_a, file_name_b):
+    """Compare objects a and b, ignoring order if the types are dicts or sortable arrays
+    """
+    aType = type(a)
+    bType = type(b)
+    if aType != bType:
+        log.warning("Mismatch: Objects have different types: %s vs %s", aType, bType)
+        return False
+
+    if aType == dict:
+        if len(a) != len(b):
+            log.warning("Mismatch: Dicts have different lengths (%s vs %s). a_keys=%s vs b_keys=%s", len(a), len(b), a.keys(), b.keys())
+            return False
+        for key in a:
+            a_value = a[key]
+            b_value = b.get(key)
+
+            if b_value is None:
+                log.warning("Mismatch: Key %s present in %s but missing from %s dict with keys %s", key, file_name_a, file_name_b, b.keys())
+                return False
+
+            if not compare_without_order(a_value, b_value, file_name_a, file_name_b):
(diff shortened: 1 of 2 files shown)


scripts/compare_snapshots.py
```diff
@@ -0,0 +1,92 @@
...
+                    "Mismatch: Key %s present in %s but missing from %s dict with keys %s",
+                    key,
+                    file_name_a,
+                    file_name_b,
+                    b.keys(),
+                )
+                return False
+
+            if not do_values_match(a_value, b_value, file_name_a, file_name_b):
+                log.warning("Mismatch: Dict values at key %s do not match", key)
```
IvanTopolcic, line 42: Maybe we could log the values here so the dev can quickly identify and fix the issue at a glance

sfirrin: The values should get logged in the recursive call to do_values_match on 41 resulting in output like

```
Performing diff of test_extension_snapshot.json against correct_extension_snapshot.json
Values AWS::Lambda::LayerVersion and AWS::Lambda::LayerVersion-BROKEN do not match
Mismatch: Dict values at key Type do not match
Mismatch: Dict values at key ProviderLevelLayerLambdaLayer do not match
Mismatch: Dict values at key Resources do not match
FAILURE: There were differences between the test_extension_snapshot.json and correct_extension_snapshot.json. Review the diff output above.
If you expected the test_extension_snapshot.json to be different generate new snapshots by running this command from a development branch on your local repository: 'UPDATE_SNAPSHOTS=true ./scripts/run_integration_tests.sh'
```

IvanTopolcic: Ah totally didn't notice this, nice work!

IvanTopolcic (approved):

PR #333, opened 2023-01-13
Continue when source code integration encounters an error.

thedavl:
### What does this PR do?

Previously if source code integration fails (such as if a repository exists in the current directory, but no git remotes are available), the plugin would return. Deployment would then continue, and the user would be left with a lambda function that has no redirect handlers set. 

### Motivation

Encountered this error when trying to deploy a test serverless app, which had a git repo initialized but had not been pushed/published to github.

### Testing Guidelines

manually tested with a serverless app that has a repo initialized in the directory, but that hasn't been published to github.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -229,7 +229,6 @@ module.exports = class ServerlessPlugin {
             });
           } catch (err) {
             this.serverless.cli.log(`Error occurred when adding source code integration: ${err}`);
-            return;
           }
         }
       }


astuyve (approved):
LGTM!

src/index.ts
```diff
@@ -229,7 +229,6 @@ module.exports = class ServerlessPlugin {
             });
           } catch (err) {
             this.serverless.cli.log(`Error occurred when adding source code integration: ${err}`);
-            return;
```
IvanTopolcic, line 232: Will something like not having git installed cause this error? If so, we may want to suppress this message and only have it for actual errors. If not, disregard!

thedavl: Afaik, if you don't have git installed then `await simpleGit.checkIsRepo()` will throw an error and the deploy will stop altogether? (though this is what would have happened previously too, simpleGit.checkIsRepo() is called outside of the try-catch).

I found [this slightly old resolved issue ](https://github.com/steveukx/git-js/pull/433) opened in the simple-git repo which states any errors (such as git not being installed) will cause the function to not return true/false, but to bubble that error up.

PR #372, opened 2023-04-03
Remove prettier usage

kimi-p:
Remove prettier usage and use relative import from `package.json` instead.


astuyve (approved):
lgtm, thanks for the quick fix!

PR #386, opened 2023-05-09
[SLS-3290] Support using DD Lambda layers published from one's own AWS account

DylanLovesCoffee:
### What does this PR do?

Adds the option `useLayersFromAccount` || `DD_USE_LAYERS_FROM_ACCOUNT` to force our plugin to use DD Lambda layers of the same name published into the executing account. Note that for this special case, we make sure the feature is _region agnostic_.

Also contains some refactoring of our code.

### Motivation

Gives support for a user to download our Lambda layer, make changes and publish their own version into their account. The versioning must be on par with our latest versions (found in layers.json)

### Testing Guidelines

Added unit tests and tested manually.

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/layer.spec.ts
+++ b/src/layer.spec.ts
@@ -713,7 +713,7 @@ describe("applyLambdaLibraryLayers", () => {
     });
   });
 
-  it("adds an tracing layer from the local AWS account of the same name", () => {
+  it("adds a tracing layer from the local AWS account of the same name", () => {
     const handler = {
       handler: { runtime: "java11" },
       type: RuntimeType.JAVA,
@@ -734,7 +734,7 @@ describe("applyLambdaLibraryLayers", () => {
     });
   });
 
-  it("adds an tracing layer from the local AWS account regardless of whether we've published to that region", () => {
+  it("adds a tracing layer from the local AWS account regardless of whether we've published to that region", () => {
     const handler = {
       handler: { runtime: "java11" },
       type: RuntimeType.JAVA,
--- a/src/layer.spec.ts
+++ b/src/layer.spec.ts
@@ -12,6 +12,7 @@ import {
   RuntimeType,
   applyLambdaLibraryLayers,
   applyExtensionLayer,
+  applyTracingLayer,
   findHandlers,
   pushLayerARN,
 } from "./layer";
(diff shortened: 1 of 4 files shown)


src/layer.spec.ts
```diff
@@ -627,6 +628,132 @@ describe("applyLambdaLibraryLayers", () => {
...
+    const mockAccountId = "123456789012";
+    const localTraceLayerARN = "arn:aws:lambda:sa-east-1:123456789012:layer:dd-trace-java:1";
+    applyTracingLayer(mockService, handler, layers, RuntimeType.JAVA, mockAccountId);
+    expect(handler.handler).toEqual({
+      runtime: "java11",
+      layers: [localTraceLayerARN],
+    });
+  });
+
+  it("adds an tracing layer from the local AWS account regardless of whether we've published to that region", () => {
```
duncanista, line 737:
```suggestion
  it("adds a tracing layer from the local AWS account regardless of whether we've published to that region", () => {
```

src/layer.ts
```diff
@@ -94,9 +94,16 @@ export function findHandlers(service: Service, exclude: string[], defaultRuntime
...
+export function applyLambdaLibraryLayers(
+  service: Service,
+  handlers: FunctionInfo[],
+  layers: LayerJSON,
+  accountId?: string,
+) {
   const { region } = service.provider;
-  const regionRuntimes = layers.regions[region];
+  // It's possible a local account layer is being used in a region we have not published to so we use a default region's ARNs
+  const shouldUseDefaultRegion = layers.regions[region] === undefined && accountId !== undefined;
```
duncanista, line 105: Just making sure I understand, we are using a fallback region `DEFAULT_REGION` to simply get a list of layers to make sure we can buid them with the users account id, assumming they have deployed them in the region they set in `provider.region|function.region`. Is this correct?

duncanista (changes requested):
Looks good, just let me know what you think of the suggestions!

src/layer.ts
DylanLovesCoffee: yup, that's exactly right 🥇

duncanista (approved):

PR #396, opened 2023-06-05
set injectLogContext to false when we use extension

zARODz11z:
### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/env.spec.ts
+++ b/src/env.spec.ts
@@ -793,6 +793,66 @@ describe("setEnvConfiguration", () => {
     ]);
   });
 
+  it("sets `DD_LOGS_INJECTION` to false when `addExtension` is true", () => {
+    const handlers: FunctionInfo[] = [
+      {
+        handler: {
+          environment: {},
+          events: [],
+        },
+        name: "function",
+        type: RuntimeType.NODE,
+      },
+    ];
+
+    setEnvConfiguration(
+      {
+        addLayers: false,
+        apiKey: "1234",
+        apiKMSKey: "5678",
+        site: "datadoghq.eu",
+        subdomain: "app",
+        logLevel: "debug",
+        flushMetricsToLogs: true,
+        enableXrayTracing: true,
+        enableDDTracing: true,
+        enableDDLogs: true,
+        subscribeToAccessLogs: true,
+        subscribeToExecutionLogs: false,
+        subscribeToStepFunctionLogs: false,
+        addExtension: true,
+        enableTags: true,
+        injectLogContext: false,
+        exclude: ["dd-excluded-function"],
+        enableSourceCodeIntegration: true,
+        uploadGitMetadata: false,
+        failOnError: false,
+        skipCloudformationOutputs: false,
+      },
+      handlers,
+    );
+
+    expect(handlers).toEqual([
+      {
+        handler: {
+          environment: {
+            DD_API_KEY: "1234",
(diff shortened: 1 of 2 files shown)


zARODz11z:
odd, I already ran update snapshots and I get 3 successes (no diffs). Something else is going on with the integ tests

sfirrin (approved):

Doxify:
Hey @zARODz11z -- would this change affect the correlation of logs with traces?

image

zARODz11z:
Hey @Doxify when using the extension we should be correlating on lambda request_id. May you give me more context? i.e. language, whether you are using the forwarder or extension, whether or not you are using the lambda layer and the version numbers?

I will try to reproduce however we might have to open a support case but I would be happy to follow up on that for you!

Doxify:
Appreciate the response, @zARODz11z! Good to know that they are being correlated based on the request id. 

Here is some more context:

- Node.js (18.x) on AWS Lambda
- We are using the following layers:
    1. Datadog-Node18-x (version 91)
    2. Datadog-Extension (version 44)

The weird thing about this is that we don't see this problem in our staging environment. On the logs page in staging, there is no problem with the trace tab (that is, each log has a trace associated with it almost immediately). However, in production, it sometimes takes 10 minutes to multiple days for logs to have a trace associated with them. For example, we have logs from last week that still don't have a trace.

I appreciate the help and no worries if you can't repro, I will open a support ticket if we can't get this figured out.

zARODz11z:
@Doxify I just setup a lambda with those settings and the correlation from logs to trace and trace to logs worked out of the box, I suspect there is another piece of the puzzle.

I made sure the changes of this pr were included in the deployment, effectively the env var DD_LOGS_INJECTION=false since are using the Datadog-extension. Therefore at the moment I don't believe this PR's changes are the culprit.

Doxify:
@zARODz11z Understood -- thanks for your help. I'll do some more digging

zARODz11z:
@Doxify Feel free to keep me posted

jeancornic:
Hi,
I have a lambda function provisioned through serverless, using the extension, and traces and logs don't correlate.
The only way I found so far to activate this is through the environment variable `DD_LOGS_INJECTION` when I modify it manually in the lambda, which can't be a long term solution.
Setting `logInjection: true` on the node js tracer directly does not seem to take precedence unfortunately.

While the request id attribute is populated on both the logs (`lambda.request_id`) and the spans (`request_id`), the matching with request id does not work, and I don't get why. I'll open a support case.

I'd like to ask: why enforcing this parameter `logInjection` to be necessarily false when using the extension? What would be the problem in having it to true if the user decides it?

zARODz11z:
> Hi @zARODz11z , I have a lambda function provisioned through serverless, using the extension, and traces and logs don't correlate. The only way I found so far to activate this is through the environment variable `DD_LOGS_INJECTION` when I modify it manually in the lambda, which can't be a long term solution. Setting `logInjection: true` on the node js tracer directly does not seem to take precedence unfortunately.
> 
> While the request id attribute is populated on both the logs (`lambda.request_id`) and the spans (`request_id`), the matching with request id does not work, and I don't get why. I'll open a support case.
> 
> I'd like to ask: why enforcing this parameter `logInjection` to be necessarily false when using the extension? What would be the problem in having it to true if the user decides it?

@duncanista @astuyve (Howdy partners)

PR #404, opened 2023-06-29
feat: Move datadog hook back to before:package:createDeploymentArtifa…

astuyve:
…cts instead of after:package:Initialize

### What does this PR do?

Had a great discussion w/ the author of [serverless-plugin-warmup](https://github.com/juanjoDiaz/serverless-plugin-warmup/issues/231) and realized we could probably fix the warmup issues with two changes:
1. Moving our hook `back` in the logical order of hooks from `after:package:initialize` to `before:package:createDeploymentArtifacts`.

This is necessary to cooperate with the `warmup` plugin which needs to inject actual code after bundlers like webpack run.

Our library only sets config, so we only need to move the hook here to accommodate `serverless-plugin-warmup`.

Reasoning from author:
```
The warmup lambdas are added on the after:package:initialize event. Which is the same event used by the datadog plugin.
However, the webpack, bundle and similar plugins use the before:package:createDeploymentArtifacts event and break things.
For that, I need to reset the warmers during the before:package:createDeploymentArtifacts event.
I wish I could get rid of that hack but then I'd break the library for all people bundling their JS.
```

2. For this to work, Datadog has to be listed _after_ `serverless-plugin-warmup`. I've added this in the readme.

### Motivation
Fully fixes our side of https://github.com/juanjoDiaz/serverless-plugin-warmup/issues/231

### Testing Guidelines
I ran this locally and tested to confirm it packages valid cloudformation for the warmer as well as my other function in the stack:
image

I manually added the `ASTUYVE` line into a local build of `serverless-plugin-warmup` to verify.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,16 @@ custom:
 
 - The Datadog Lambda Extension is now the default mechanism for transmitting telemetry to Datadog.
 
+## Working with serverless-plugin-warmup
+This library is compatible at best effort with [serverless-plugin-warmup](https://github.com/juanjoDiaz/serverless-plugin-warmup). If you want to exclude the warmer function from Datadog, you can dot hat with the `exclude` feature of this library.
+
+To properly package your application this plugin *must* be listed _after_ `serverless-plugin-warmup` in your `serverless.yml` file:
+```yaml
+plugins:
+  - serverless-plugin-warmup
+  - serverless-plugin-datadog
+```
+
 ## Opening Issues
 
 If you encounter a bug with this package, let us know by filing an issue! Before opening a new issue, please search the existing issues to avoid duplicates.
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ custom:
 ## Working with serverless-plugin-warmup
 This library is compatible at best effort with [serverless-plugin-warmup](https://github.com/juanjoDiaz/serverless-plugin-warmup). If you want to exclude the warmer function from Datadog, use the `exclude` feature of this library.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -212,6 +212,16 @@ custom:
 
 - The Datadog Lambda Extension is now the default mechanism for transmitting telemetry to Datadog.
 
+## Working with serverless-plugin-warmup
+This library is compatible at best effort with [serverless-plugin-warmup](https://github.com/juanjoDiaz/serverless-plugin-warmup). If you want to exclude the warmer function from Datadog, you can dot hat with the `exclude` feature of this library.
```
aliciascott, line 216:
```suggestion
This library is compatible at best effort with [serverless-plugin-warmup](https://github.com/juanjoDiaz/serverless-plugin-warmup). If you want to exclude the warmer function from Datadog, use the `exclude` feature of this library.
```

README.md
```diff
@@ -212,6 +212,16 @@ custom:
...
+
+To properly package your application this plugin *must* be listed _after_ `serverless-plugin-warmup` in your `serverless.yml` file:
```
aliciascott, line 218:
```suggestion
To properly package your application, this plugin *must* be listed _after_ `serverless-plugin-warmup` in your `serverless.yml` file:
```

aliciascott (changes requested):
A couple small changes in the readme

aliciascott (approved):
Looks good thanks!

sfirrin (approved):

PR #413, opened 2023-07-21
Add `enableStepFunctionsTrace` option to replace `subscribeToStepFunctionLogs`

kimi-p:
https://datadoghq.atlassian.net/browse/SLS-3823

- Add `enableStepFunctionsTrace` option to have what `subscribeToStepFunctionLogs` and add `DD_TRACE_ENABLED=true` tag on the state machine.
- Remove `subscribeToStepFunctionLogs` option from README (not removing the logic in code since a couple private beta users are still using it)

## Tested in sandbox account

image

image

image

## Not able to add warning message on Lambda basic legacy integration
❌ Add warning message on the basic legacy lambda integration steps when `mergeStepFunctionAndLambdaTraces` is set to true since merging does not work with the legacy lambda integrations.

This does not work because JSON file generated by Serverless Framework use a hash to point to a resource on the step's Resource field, not using lambda arn directly as we see on the AWS UI.

```

## legacy integration step
\"LegacyLambda\":{\"Type\":\"Task\",\"Resource\":\"${dcf1c5b76ae09b9dc49ad82d5ef8290d}\",\"End\":true}}

## full definitiion
"DefinitionString": {
          "Fn::Sub": [
            "{\"Comment\":\"Example Step Function with preconfigured log group subscribed to using the Datadog Serverless Plugin\",\"StartAt\":\"ChoiceStep\",\"States\":{\"ChoiceStep\":{\"Type\":\"Choice\",\"Comment\":\"some choice to make\",\"Choices\":[{\"Variable\":\"$.Comment\",\"StringEquals\":\"Insert your JSON here\",\"Next\":\"InvokeLambda\"}]},\"InvokeLambda\":{\"Type\":\"Task\",\"Resource\":\"arn:aws:states:::lambda:invoke\",\"Parameters\":{\"FunctionName\":\"${dcf1c5b76ae09b9dc49ad82d5ef8290d}\",\"Payload.$\":\"States.JsonMerge($$, $, false)\"},\"Next\":\"LegacyLambda\"},\"LegacyLambda\":{\"Type\":\"Task\",\"Resource\":\"${dcf1c5b76ae09b9dc49ad82d5ef8290d}\",\"End\":true}}}",
            {
              "dcf1c5b76ae09b9dc49ad82d5ef8290d": {
                "Fn::GetAtt": [
                  "StepfunctionsDashtestingDashfunctionLambdaFunction",
                  "Arn"
                ]
              }
            }
          ]
        }
```


src/forwarder.ts
```diff
@@ -142,15 +142,22 @@ export async function addStepFunctionLogGroup(aws: Aws, resources: any, stepFunc
...
-    stateMachineObj.Properties.Tags = [];
-  }
   stateMachineObj.Properties?.Tags?.push({
     Key: "dd_sls_plugin",
     Value: `v${version}`,
   });
 }
 
+export function addDdTraceEnabledTag(stateMachineObj: any, enableStepFunctionsTrace: undefined | boolean) {
+  if (!enableStepFunctionsTrace) {
```
nine5two7, line 152: So the logic is: Unless the user specifically set enableStepTunctionsTrace to be false, DD_Trace_Enabled will always be set true, and step function tracing is enabled. Right?

kimi-p: I think I missed the logic where we set this flag default to true. And I'll need to catch all the possible exceptions for this logic in case there are any errors that may block any customers' deployment.

nine5two7 (approved):
LGTM

README.md
```diff
@@ -64,7 +63,7 @@ To further configure your plugin, use the following custom parameters in your `s
...
 | `mergeStepFunctionAndLambdaTraces` | When set to `true`, Lambda traces merge with the calling Step Functions trace. Defaults to `false`.                                                                                                                                                                                                                                                                                                                          |
-
+| `enableStepFunctionsTrace`    | Enable automatic subscription of the Datadog Forwarder to Step Function log groups and Step Functions tracing. If no Step Function log groups are configured then they will be automatically created. Requires setting `forwarderArn`. Defaults to `false`.                                                                                                                                                                  |
```
urseberry, line 66:
```suggestion
| `enableStepFunctionsTrace`    | Enable automatic subscription of the Datadog Forwarder to Step Function log groups and Step Functions tracing. If no Step Function log groups are configured, then they are automatically created. Requires setting `forwarderArn`. Defaults to `false`.                                                                                                                                                                  |
```

src/forwarder.ts
kimi-p: I just realized that we cannot set this flag to true. So there is no need to add try catch for it then.

urseberry (approved):

nine5two7 (approved):
LGTM

PR #433, opened 2023-10-11
Search monitor patch

johnbuildsthings:
### What does this PR do?
This pr adds pagination to the seachMonitors method.

### Motivation

We encountered an error where a serverless stack with 33 monitors would always return a 400 error while deploying monitors unless we first manually deleted all monitors associated with the stack.  Upon investigation I found that the searchMonitors method only fetches 30 monitors (the page size default for the search api), thus ensuring that there will always be 3 monitors that "exist" but are not part of the [serverlessMonitorIdByMonitorId](https://github.com/DataDog/serverless-plugin-datadog/blob/master/src/monitors.ts#L163) map.  DD will attempt to create instead of update these monitors resulting in a 400 error..

### Testing Guidelines

I added an additional test case which will return metadata stating that more than 1 page of monitors exists, this will trigger the searchMonitors method to iterate until it retrieves all the monitors.  No existing monitors are failing. I updated one of the existing test cases to return and array of monitors instead of a monitor object, this is what's expected base on reading the code and experiments with the api.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/monitor-api-requests.spec.ts
+++ b/src/monitor-api-requests.spec.ts
@@ -170,37 +170,39 @@ describe("searchMonitor", () => {
       status: 200,
       json: () => ({
         metadata: { total_count: 3, page: 0, per_page: 30, page_count: 1 },
-        monitors: [{
-          status: "No Data",
-          scopes: ["aws_cloudformation_stack-id:cloudformation_stack_id"],
-          classification: "metric",
-          creator: {
-            handle: "datadog@datadoghq.com",
+        monitors: [
+          {
+            status: "No Data",
+            scopes: ["aws_cloudformation_stack-id:cloudformation_stack_id"],
+            classification: "metric",
+            creator: {
+              handle: "datadog@datadoghq.com",
+              id: 1234567,
+              name: "H Jiang",
+            },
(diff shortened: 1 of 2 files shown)


sfirrin:
Thanks a lot for this @johnbuildsthings. Code looks good to me

The branch is failing the linter check, I think that should be fixed by running `prettier . write`

It's failing the integration test because the GH action doesn't get environment variables with community PRs, I'm going to run them in a duplicate testing branch

sfirrin:
Integration tests pass https://github.com/DataDog/serverless-plugin-datadog/actions/runs/6643218659/job/18049787998, just need the lint and this should be ready

johnbuildsthings:
@sfirrin I fixed the lint errors.  Let me know if there is more that I need to do to keep this moving?

sfirrin (approved):

sfirrin:
That should be all @johnbuildsthings, thanks again for the contribution!

Merged just now and will release a new version with this today

sfirrin:
Released in https://github.com/DataDog/serverless-plugin-datadog/releases/tag/v5.49.0

Thanks @johnbuildsthings!

PR #441, opened 2023-11-06
[SLS-4080] Replace hardcoded monitor definitions 

hannahqjiang:
### What does this PR do?

This PR: 
* removes the hardcoded monitor definitions that for serverless recommended monitors 
* queries the monitor API endpoint for recommended monitor definitions instead (see `getRecommendedMonitors()` function) 
* adds unit tests for `getRecommendedMonitors()` 
* updates existing unit tests to use mock values for recommended monitors 
 

### Motivation

https://datadoghq.atlassian.net/browse/SVLS-4080?atlOrigin=eyJpIjoiOWIxMDBiZmE3YzExNGVmNDk3NjFjMTcxNTk4MTYyNWQiLCJwIjoiaiJ9
### Testing Guidelines

* Added/changed unit tests 
* deployed a test lambda function with a test monitor defined in the YAML file 

Here's an example for a recommended monitor that you can add to your `serverless.yml` file for testing:

```
 monitors:
   - out_of_memory:
            options:
              thresholds:
                critical: 3
            message: "At least one invocation in the selected time range ran out of memory."
            tags:
              - "example_monitor"
```

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


src/monitors.spec.ts
```diff
@@ -161,29 +162,49 @@ const TIMEOUT_MONITOR_PARAMS = {
...
   it("returns valid monitor params for an updated custom monitor", async () => {
-    const monitorParams = buildMonitorParams(UPDATED_CUSTOM_MONITOR_2, "cloud_formation_id", "service", "env");
+    const monitorParams = buildMonitorParams(UPDATED_CUSTOM_MONITOR_2, "cloud_formation_id", "service", "env", RECOMMENDED_MONITORS);
     expect(monitorParams).toEqual(UPDATED_CUSTOM_MONITOR_2_PARAMS);
   });
   it("returns valid monitor params for Increased Cost monitor", async () => {
-    const monitorParams = buildMonitorParams(INCREASED_COST_MONITOR, "cloud_formation_id", "service", "env");
-    expect(monitorParams).toEqual(INCREASED_COST_MONITOR_PARAMS);
+    // const monitorParams = buildMonitorParams(INCREASED_COST_MONITOR, "cloud_formation_id", "service", "env", RECOMMENDED_MONITORS);
+    // expect(monitorParams).toEqual(INCREASED_COST_MONITOR_PARAMS);
```
hannahqjiang, line 204:
```suggestion

```

src/monitors.ts
```diff
@@ -1,5 +1,11 @@
-import { SERVERLESS_MONITORS } from "./serverless_monitors";
-import { updateMonitor, createMonitor, deleteMonitor, getExistingMonitors } from "./monitor-api-requests";
+// import { SERVERLESS_MONITORS } from "./serverless_monitors";
```
hannahqjiang, line 1:
```suggestion

```

src/monitor-api-requests.ts
```diff
@@ -138,3 +151,44 @@ export async function getExistingMonitors(
...
+  const recommendedMonitorsData = json.data;
+  recommendedMonitorsData.forEach((recommendedMonitorParam: RecommendedMonitorParams) => {
+    const recommendedMonitor: ServerlessMonitor = {
+      name: recommendedMonitorParam.attributes.name,
+      threshold: recommendedMonitorParam.attributes.options.thresholds.critical,
+      message: recommendedMonitorParam.attributes.message,
+      type: recommendedMonitorParam.attributes.type,
+      query: (cloudFormationStackId: string, criticalThreshold: number) => {
+        let query = recommendedMonitorParam.attributes.query;
+        query = query.replace(/\$scope*/g, `aws_coudformation_stack-id:${cloudFormationStackId}`);
```
hannahqjiang, line 180:
```suggestion
// replace $scope with cloudformation_stack_id
        query = query.replace(/\$scope*/g, `aws_coudformation_stack-id:${cloudFormationStackId}`);
```

clifordshelton (commented):
Our code for creating and getting enabled monitors uses the tag `serverless_monitor_id` as seen in below screenshot..
image

But the newly defined monitors use the tag `serverless_id` like in [this example](https://github.com/DataDog/integrations-internal-core/blob/0407c8048938cb1bfd865ab3128ec561cbbc7930/serverless/assets/monitors/high_cold_start_rate.json#L14)
So our code to check for already enabled monitors might not work and we'd have to do some changes for it

src/monitor-api-requests.ts
```diff
@@ -32,7 +46,7 @@ export async function createMonitor(
     },
     body: JSON.stringify(monitorParams),
   });
-
+  console.log(JSON.stringify(monitorParams));
```
hannahqjiang, line 49:
```suggestion
```

clifordshelton:
- I created a monitor using the plugin with this definition...
```
    monitors:
      - out_of_memory:
          options:
            thresholds:
              critical: 2
          message: "More than 2 invocations ran out of memory."
          tags:
            - "example_monitor:true"
```
- The monitor got created successfully
image
- But when I view the [lambda function's details](https://dd.datad0g.com/functions?search=functionname%3Apython-plugin-test-dev-api&cloud=aws&entity_view=lambda_functions&panel_end=1701456040999&panel_paused=true&panel_start=1701452440999&panel_tab=configuration&selection=aws-lambda-functions%2Bpython-plugin-test-dev-api%2Bus-east-1%2B425362996713&start=1701452440999&end=1701456040999&paused=true), the monitor shows up unenabled
image

src/monitor-api-requests.ts
```diff
@@ -130,11 +142,53 @@ export async function getExistingMonitors(
...
+  recommendedMonitorsData.forEach((recommendedMonitorParam: RecommendedMonitorParams) => {
+    const recommendedMonitor: ServerlessMonitor = {
+      name: recommendedMonitorParam.attributes.name,
+      threshold: recommendedMonitorParam.attributes.options.thresholds.critical,
+      message: recommendedMonitorParam.attributes.message,
+      type: recommendedMonitorParam.attributes.type,
+      query: (cloudFormationStackId: string, criticalThreshold: number) => {
+        let query = recommendedMonitorParam.attributes.query;
+        // replace $scope with cloudformation_stack_id
+        query = query.replace(/\$scope*/g, `aws_coudformation_stack-id:${cloudFormationStackId}`);
```
clifordshelton, line 180: I just noticed that the tag says aws_**coud**formation_stack-id! 😄 .. Glad I caught that or else these monitors would never work! Once you have some more changes for the comments, I'm going to try creating a monitor and see if it really alerts so that we can be sure they're working..

src/monitors.ts
```diff
@@ -46,23 +68,23 @@ export function buildMonitorParams(monitor: Monitor, cloudFormationStackId: stri
...
   ];
 
-  if (checkIfRecommendedMonitor(serverlessMonitorId)) {
-    let criticalThreshold = SERVERLESS_MONITORS[serverlessMonitorId].threshold;
+  if (isRecommendedMonitor(serverlessMonitorId, recommendedMonitors)) {
+    let criticalThreshold = recommendedMonitors[serverlessMonitorId].threshold;
+
     if (monitorParams.options) {
       if (monitorParams.options.thresholds) {
         if (monitorParams.options.thresholds.critical) {
```
clifordshelton, line 76: Two comments here
1) The last check is `if (monitorParams.options.thresholds.critical)` which will be false when the value for critical threshold is 0 in which case the custom threshold will not get picked up. This check should instead be `if (monitorParams.options.thresholds.critical !== undefined)`

2) This being a typescript file, instead of having 3 nested if conditions, this can be simplified to...
```
if (monitorParams.options?.thresholds?.critical !== undefined) {
```

src/monitors.ts
```diff
@@ -73,8 +95,8 @@ export function buildMonitorParams(monitor: Monitor, cloudFormationStackId: stri
  * @param serverlessMonitorId - Unique ID string defined for each monitor
  * @returns true if a given monitor is a serverless recommended monitor
  */
-function checkIfRecommendedMonitor(serverlessMonitorId: string) {
-  return Object.keys(SERVERLESS_MONITORS).includes(serverlessMonitorId);
+function isRecommendedMonitor(serverlessMonitorId: string, recommendedMonitors: RecommendedMonitors) {
+  return Object.keys(recommendedMonitors).includes(serverlessMonitorId);
```
clifordshelton, line 99: You could simply do the below..
```suggestion
  return recommendedMonitors[serverlessMonitorId] !== undefined;
```
Executing `object.keys` and then running `.includes` is a roundabout way of doing what you're trying to do because it has to do two iteration operations..
1) `object.keys` should build and return an array of all the key names
2) `.includes` operation would have to iterate through the values of that array and string match each value against the ID

Whereas the suggested change will simply use the direct object key lookup that's the nature of a JSON object and is much more efficient

src/monitor-api-requests.spec.ts
```diff
@@ -388,3 +394,322 @@ describe("searchMonitor", () => {
...
+          type: "query alert",
+          query: (cloudFormationStackId: string, criticalThreshold: number) => {
+            return `sum(last_15m):sum:aws.lambda.enhanced.invocations{cold_start:true,${cloudFormationStackId}} by {aws_account,functionname,region}.as_count() / sum:aws.lambda.enhanced.invocations{${cloudFormationStackId}} by {aws_account,functionname,region}.as_count() >= ${criticalThreshold}`;
```
clifordshelton, line 675: I'm wondering if this unit test caught [the typo](https://github.com/DataDog/serverless-plugin-datadog/pull/441/files#r1412523334) of `coudFormation` instead of `cloudFormation` or not.. Could you check and make sure it's working as intended?

src/monitor-api-requests.ts
hannahqjiang: Oops, good catch 😅

src/monitor-api-requests.spec.ts
hannahqjiang: I think because cloudFormation is passed in as a param for a function, it returns [Function: query] instead of the actual query string. Here's what an example mocked response would look like: 

```
timeout: {
        name: 'Timeout on $functionName in $regionName for $awsAccount',
        threshold: 1,
        message: 'At least one invocation in the selected time range timed out. This occurs when your function runs for longer than the configured timeout or the global Lambda timeout. Resolution: [Distributed tracing](https://docs.datadoghq.com/serverless/distributed_tracing) can help you pinpoint slow requests to APIs and other microservices. You can also consider increasing the timeout of your function. Note that this could affect your AWS bill.',
        type: 'query alert',
        query: [Function: query]
      }
    }
```

src/monitor-api-requests.ts
```diff
@@ -177,7 +177,8 @@ export async function getRecommendedMonitors(site: string, monitorsApiKey: strin
       query: (cloudFormationStackId: string, criticalThreshold: number) => {
         let query = recommendedMonitorParam.attributes.query;
         // replace $scope with cloudformation_stack_id
-        query = query.replace(/\$scope*/g, `aws_coudformation_stack-id:${cloudFormationStackId}`);
+        query = query.replace(/\$scope/g, `aws_cloudformation_stack-id:${cloudFormationStackId}`);
+        // console.log(query);
```
clifordshelton, line 181: 🧹 To be removed..

package.json
```diff
@@ -58,6 +58,7 @@
   },
   "dependencies": {
     "@datadog/datadog-ci": "^2.17.2",
+    "glob": "^10.3.10",
```
clifordshelton, line 61: 🧹  To be removed

.yarnrc.yml
```diff
@@ -0,0 +1 @@
+nodeLinker: node-modules
```
clifordshelton, line 1: 🧹  To be removed

clifordshelton (commented):
🧹  Are the changes in `install-state.gz` and `yarn.lock` files expected?

clifordshelton (approved):
Awesome! Great work Hannah! 👍

PR #455, opened 2024-01-02
Add logic to suppress duplicate messages

clifordshelton:
### What does this PR do?

Includes changes that are meant to suppress duplicate log lines in SLS Plugin output

### Motivation

When using SLS plugin, we noticed that there are duplicate messages in the output which might seem a bit confusing to users. We wanted to suppress these messages (Ideally we would want the duplication to not happen instead of simply suppressing the messages but that's not fully possible as mentioned in the Jira ticket)
Jira Ticket: https://datadoghq.atlassian.net/browse/SVLS-3544

image

### Testing Guidelines

Try running the plugin and see if these messages show only once or multiple times

image

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -91,8 +91,8 @@ module.exports = class ServerlessPlugin {
   };
   constructor(private serverless: Serverless, private options: Serverless.Options) {}
 
-  private displayedMessages:{[msg:string]: true} = {};
-  private logToCliOnce (message: string) {
+  private displayedMessages: { [msg: string]: true } = {};
+  private logToCliOnce(message: string) {
     if (this.displayedMessages[message] === undefined) {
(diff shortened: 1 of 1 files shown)


duncanista:
> Do we know why it got duplicated? More specifically, does the duplicated logs actually indicate duplicated invocations? Asking just to make sure we fix the root cause.

@joeyzhao2018 it happens because some hooks might not be executed on every case. For example, there's scenarios where only certain `after:` hooks will be executed, but not all of them. And viceversa, only one `after:` hook might execute. Therefore, the solution of suppressing messages! Let me know if you want more details about this.

duncanista (approved):

Issue #480, opened 2024-04-01
Plugin configuration with only log forwarding (via forwarder) and monitors should not require wrapper


max-lobur:
## Expected Behavior
Extensions/Layers/Library should not be not required when using plugin only to link log forwarder and setup monitors.

## Actual Behavior
Lambda error:
```
/opt/datadog_wrapper: does not exist
```

## Steps to Reproduce the Problem
```
  datadog:
    enabled: true
    site: us5.datadoghq.com
    apiKey: <>
    appKey: <>
    enableDDLogs: false
    enableXrayTracing: false
    enableDDTracing: false
    enableASM: false
    enableProfiling: false
    enableSourceCodeIntegration: false
    enableColdStartTracing: false
    uploadGitMetadata: false
    addLayers: false
    addExtension: false
    encodeAuthorizerContext: false
    decodeAuthorizerContext: false
    subscribeToAccessLogs: true
    subscribeToExecutionLogs: true
    forwarderArn: <arn>
    failOnError: true
    logLevel: DEBUG
    service: api
    env: dev
    monitors:
      - high_error_rate: {}
      - timeout: {}
      - high_throttles: {}
```

```
sls deploy
```

## Specifications

  - Serverless Framework version: 
Framework Core: 3.38.0 (local) 3.38.0 (global)
Plugin: 7.2.0
SDK: 4.5.1
  - Datadog Serverless Plugin version: 5.60.0
  - Lambda function runtime (Python 3.7, Node 10, etc.): python3.10
  
## Stacktrace
  
  ```
/opt/datadog_wrapper: does not exist
  ```

astuyve:
Hi @max-lobur, thanks for reaching out!

Can you remove `enableASM: false` and try again?

I'll reach out to that group and ask them review a change to fix this logic.
I'm sorry about this.

Thanks!
AJ

max-lobur:
@astuyve with `enableASM: false` I'm getting
```
No module named 'datadog_lambda'
```
Guess it's not over yet

max-lobur:
let me know if you need the new issue

astuyve:
Hi Max, sorry about that, I forgot GitHub closes issues when the corresponding PR is merged.

For now, one more thing to try is setting `customHandler: your-handler.handler`

[...]

PR #482, opened 2024-04-02
Feat: Support optionally disabling function handler redirection

astuyve:
### What does this PR do?

1. Finishes #480 - Supports users who are not using a tracing library at all, but still want to use this library to forward logs or create monitors
2. More easily use custom installation method where users call `Datadog()` methods explicitly in their handler code and thus don't have to set `customHandler`, which only applies globally to all functions in the template.
3. Supports users who want to send custom metrics using the library as a pure library, and not as an APM-enabled handler.

### Motivation
#480

### Testing Guidelines
Manually tested in our serverless sandbox environment using a python 3.12 function. Also automated tests.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


alai97 (approved):

duncanista (approved):

purple4reina (commented):
🚢

PR #499, opened 2024-07-22
Fix unresolved template variables in recommended monitor titles

lym953:
### Motivation

To fix a bug that emails triggered by Serverless recommended variables contain unresolved template variables in the title, e.g. 
```
Triggered: High Error Rate on $functionName in $regionName for $awsAccount on aws_account:425362996713,functionname:yiming-serverless-demo-dev-yiming_hello,region:sa-east-1
```

image

The monitor name is also incorrect on Datadog UI:

Screenshot 2024-07-22 at 12 48 20 PM
where, for example, `$functionName` should be `{{functionname.name}}`.

### What does this PR do?

replacing the dollar-formed template variables by their curly-braces-formed default values as defined in the recommended monitor JSON, e.g. replacing `$functionName` by `{{functionname.name}}`.

##### Steps
1. In the Lambda handler of a Serverless project, make the function fail:
```
def yiming_hello(event, context):
    raise Exception('Let it fail!')
```
2. Add this to `serverless.yml` of a Serverless project:
```
    monitors:
      - high_error_rate:
          tags: ["team:serverless-one", "service:yiming-serverless-demo"]
          message: "More than 10% of the function’s invocations were errors in the selected time range. Notify @yiming.luo@datadoghq.com"
```
3. Deploy the changes to Serverless sandbox AWS account, following [this internal wiki](https://datadoghq.atlassian.net/wiki/spaces/SLS/pages/2722794789/Datadog+Serverless+Plugin#Local-Changes)
4. Check the monitor on Serverless UI. Check emails.

##### Results
Before: see "motivation" section.
After:
Screenshot 2024-07-22 at 2 07 58 PM
Screenshot 2024-07-22 at 12 43 00 PM

#### 2. Automated tests
```
npm test monitors.spec.ts
npm test monitor-api-requests.spec.ts
```

### Additional Notes

This PR assumes that every Serverless recommended monitor has template variables, at least this is true for all the 7 recommended monitors at the moment. Please let me know if this assumption is wrong.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Misc (docs, refactoring, dependency upgrade, etc.)

### Check all that apply

- [x] This PR's description is comprehensive
- [ ] This PR contains breaking changes that are documented in the description
- [ ] This PR introduces new APIs or parameters that are documented and unlikely to change in the foreseeable future
- [ ] This PR impacts documentation, and it has been updated (or a ticket has been logged)
[...]

PR #500, opened 2024-07-25
Handle edge cases around template variables in recommended monitor title

lym953:
### Motivation
Right now the code assumes that:
1. every recommended monitor has template variables, i.e. `template_variable` field is set in the JSON file
2. every template variable has a default value, i.e. the `default` field is an non-empty list

However,
- 1 will soon be false in that a recommended monitor with no default value is being added https://github.com/DataDog/integrations-internal-core/pull/806
- 2 is usually true, but it will be false if a monitor is misconfigured.

In either case, Serverless plugin code will throw an exception, and recommended monitors will fail to be created.

### What does this PR do?

Gracefully handle these two cases by:
1. not assuming that every recommended monitor has template variables.
2. If a template variable has no default value, then ignore it in interpolation code.

### Testing Guidelines

1. automated test for the new behavior:
```
npm test src/monitor-api-requests.spec.ts
npm test src/monitors.spec.ts
```
2. manual test: same as #499, to ensure that existing behavior is not broken. A recommended monitor is created successfully.

### Additional Notes

Thanks @TalUsvyatsky for guidance and @clifordshelton for pointing out these two cases.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/monitors.spec.ts
+++ b/src/monitors.spec.ts
@@ -77,6 +77,10 @@ const DEFAULT_TIMEOUT_MONITOR: Monitor = {
   timeout: {},
 };
 
+const NO_TEMPLATE_VARIABLE_MONITOR: Monitor = {
+  test_no_template_variable: {},
+};
+
 const CUSTOM_MONITOR_1_PARAMS = {
   name: "Custom Monitor 1",
   query: "max(next_1w):forecast(avg:system.load.1{*}, 'linear', 1, interval='60m', history='1w', model='default') >= 3",
@@ -171,8 +175,23 @@ const TIMEOUT_MONITOR_PARAMS = {
   message:
     "At least one invocation in the selected time range timed out. This occurs when your function runs for longer than the configured timeout or the global Lambda timeout. Resolution: [Distributed tracing](https://docs.datadoghq.com/serverless/distributed_tracing) can help you pinpoint slow requests to APIs and other microservices. You can also consider increasing the timeout of your function. Note that this could affect your AWS bill.",
 };
+const NO_TEMPLATE_VARIABLE_MONITOR_PARAMS = {
+  name: "This is to ensure that serverless plugin works properly when a recommended monitor has no template variable",
+  query: "false",
+  tags: [
+    "serverless_monitor_type:single_function",
+    "serverless_monitor_id:test_no_template_variable",
+    "aws_cloudformation_stack-id:cloud_formation_id",
+    "created_by:dd_sls_plugin",
+    "env:env",
+    "service:service",
+  ],
+  options: {},
+  type: "metric alert",
+  message: "This alert is not supposed to be triggered.",
+};
 const DEFAULT_TIMEOUT_MONITOR_PARAMS = {
-  name: "Timeout on {{functionname.name}} in {{region.name}} for {{aws_account.name}}",
+  name: "Timeout on {{functionname.name}} in {{region.name}} for {{aws_account.name}} with $varNoDefault",
   query:
     "avg(last_15m):sum:aws.lambda.duration.maximum{aws_cloudformation_stack-id:cloud_formation_id} by {aws_account,functionname,region}.as_count() / (sum:aws.lambda.timeout{aws_cloudformation_stack-id:cloud_formation_id} by {aws_account,functionname,region}.as_count() * 1000) >= 1",
   tags: [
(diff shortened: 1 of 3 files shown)


src/monitors.ts
```diff
@@ -244,16 +246,23 @@ export function replaceCriticalThreshold(query: string, criticalThreshold: numbe
...
+ * @param recommendedMonitors  - recommended monitors
  * @param name - The string that may contain uninterpolated template variables,
  *        e.g. "High Error Rate on $functionName in $regionName for $awsAccount"
  * @returns Interpolation result, e.g. "High Error Rate on {{functionname.name}} in {{region.name}} for {{aws_account.name}}"
  */
-function interpolateTemplateVariables(templateVariables: TemplateVariable[], name: string): string {
-  templateVariables.forEach((templateVariable) => {
-    name = name.replace("$" + templateVariable.name, templateVariable.defaults[0]);
-  });
+function interpolateTemplateVariables(recommendedMonitor: ServerlessMonitor, name: string): string {
```
TalUsvyatsky, line 258: non-blocking nit: should we call the parameter something like `baseString` instead of `name` in case we later want to interpolate other parts of the monitor, such as the `message` field?

src/monitors.spec.ts
```diff
@@ -237,6 +246,14 @@ const RECOMMENDED_MONITORS: RecommendedMonitors = {
     },
     templateVariables: TEMPLATE_VARIABLES,
   },
+  test_no_template_variable: {
```
TalUsvyatsky, line 249: Should we have a test that calls `buildMonitorParams` with a `test_no_template_variable` monitor and verifies that the `name` is unchanged?

lym953: Good point!

src/monitors.ts
lym953: I had thought about a better name for "name" but couldn't find one. Thanks for the suggestion!

lym953 (commented):
Addressed @TalUsvyatsky 's comments

TalUsvyatsky (approved):
Nice job, looks good!

src/monitors.ts
```diff
@@ -84,7 +85,7 @@ export function buildMonitorParams(
       monitorParams.message = recommendedMonitor.message;
     }
     if (!monitorParams.name) {
-      monitorParams.name = interpolateTemplateVariables(recommendedMonitor.templateVariables, recommendedMonitor.name);
+      monitorParams.name = interpolateTemplateVariables(recommendedMonitor, recommendedMonitor.name);
```
clifordshelton, line 88: **Nit:** It seems odd that we're both passing the entire recommended monitor as well as the name in a separate param. This function could simply take the recommended monitor alone and read the name from it.

**Another Nit:** With the changes done, a name such as `getInterpolatedMonitorName` might be easier to understand than `interpolateTemplateVariables`

src/monitors.ts
```diff
@@ -244,16 +246,24 @@ export function replaceCriticalThreshold(query: string, criticalThreshold: numbe
...
-function interpolateTemplateVariables(templateVariables: TemplateVariable[], name: string): string {
-  templateVariables.forEach((templateVariable) => {
-    name = name.replace("$" + templateVariable.name, templateVariable.defaults[0]);
-  });
-  return name;
+function interpolateTemplateVariables(recommendedMonitor: ServerlessMonitor, baseString: string): string {
+  const templateVariables = recommendedMonitor.templateVariables;
+  let interpolatedString = baseString;
+  if (templateVariables !== undefined && templateVariables.length > 0) {
+    templateVariables.forEach((templateVariable) => {
```
clifordshelton, line 262: ***Nit:*** Just a small suggestion here. Since this is typescript, this code can be simplifed to just the below
```
    let interpolatedString = recommendedMonitor.name;
    recommendedMonitor.templateVariables?.forEach((templateVariable) => {
```

src/monitors.ts
```diff
@@ -244,16 +246,24 @@ export function replaceCriticalThreshold(query: string, criticalThreshold: numbe
...
+      if (templateVariable.defaults.length > 0) {
+        interpolatedString = interpolatedString.replace("$" + templateVariable.name, templateVariable.defaults[0]);
+      }
```
clifordshelton, line 265: I'm curious now 🤔 . Why does the monitors team have defaults defined as an array for a template variable? What does it mean when there are multiple values in this array? We are always taking just the first value and consuming it. It would be better to understand their intention so that we handle it appropriately.

lym953: Good point. I guess an empty array means no default value, an array with one element means default value exists, and all the elements after the first one are ignored.

Confirming with them:
https://datadoghq.atlassian.net/wiki/spaces/AP/pages/3230303531/Product-based+Recommended+Monitors+Usage+Guide?focusedCommentId=3948118079

PR #521, opened 2024-09-19
chore: Simplify index signature iteration code

lym953:
#### 1. Simplify the code to iterate over an index signature.

For example, for a TypeScript index signature:
```
{ [key: string]: GeneralResource }
```

Existing iteration code:
```
for (const resourceName in resources) {
  if (resources.hasOwnProperty(resourceName)) {
    const resourceObj: GeneralResource = resources[resourceName];
    // do something
  }
}
```

This PR changes it to:
```
for (const [resourceName, resourceObj] of Object.entries(resources)) {
  // do something
}
```

I'm removing `resources.hasOwnProperty()` because I think it should be always true. Correct me if I'm wrong.

#### 2. Change a if-check to an early break
Change
```
for (...) {
  if (A) {
    if (B) {
      // do something
    }
  }
}
```
to
```
for (...) {
  if (!A || !B) {
    continue
  }
  // do something
}
```

### Motivation

To simplify code.

#### 1. Automated testing
Passed all existing tests by running `npm test`

##### Steps
1. Write `serverless.yml`. Partial content:
```
functions:
  yiming_plugin_hello:
    handler: handler.yiming_plugin_hello

stepFunctions:
  stateMachines:
    yimingServerlessPluginTestStateMachine:
      name: YimingServerlessPluginTestStateMachine
      definition:
        Comment: "A Hello World example of the Amazon States Language using an AWS Lambda Function"
        StartAt: PrepareDataPass
        States:
          PrepareDataPass:
            Type: Pass
            Result:
              message: "Data prepared successfully"
            Next: InvokeLambdaStep
          InvokeLambdaStep:
            Type: Task
            Resource: arn:aws:states:::lambda:invoke
            Parameters:
              FunctionName:
                Fn::GetAtt: [yiming_plugin_hello, Arn]
              Payload.$: $
            End: true
  validate: true

plugins:
  - serverless-step-functions
  - serverless-plugin-datadog

custom:
  datadog:
    site: datadoghq.com
    apiKey: ...
    appKey: ...
    propagateUpstreamTrace: true
    enableDDTracing: true
    enableDDLogs: true
    subscribeToAccessLogs: true
    forwarderArn: ...
    enableStepFunctionsTracing: true
```
2. `serverless deploy`

##### Result
1. Deployment was successfully
2. Go to AWS console UI. The State Machine definition has been updated to include `"Payload.$": "States.JsonMerge($$, $, false)"`

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/span-link.ts
+++ b/src/span-link.ts
@@ -5,16 +5,12 @@ export function mergeStepFunctionAndLambdaTraces(
   resources: { [key: string]: GeneralResource },
   serverless: Serverless,
 ): void {
-  for (const resourceName in resources) {
-    if (resources.hasOwnProperty(resourceName)) {
-      const resourceObj: GeneralResource = resources[resourceName];
-      if (resourceObj.Type === "AWS::StepFunctions::StateMachine") {
-        if (resourceObj.Properties) {
-          const definitionString = resourceObj.Properties?.DefinitionString!;
-          const newDefString = updateDefinitionString(definitionString, serverless, resourceName);
-          resourceObj.Properties.DefinitionString = newDefString;
-        }
-      }
+  for (const [resourceName, resourceObj] of Object.entries(resources)) {
+    if (resourceObj.Type !== "AWS::StepFunctions::StateMachine" || !resourceObj.Properties) {
+      continue;
     }
+    const definitionString = resourceObj.Properties?.DefinitionString!;
+    const newDefString = updateDefinitionString(definitionString, serverless, resourceName);
+    resourceObj.Properties.DefinitionString = newDefString;
   }
 }
--- a/src/step-functions-helper.ts
+++ b/src/step-functions-helper.ts
@@ -126,33 +126,30 @@ export function updateDefinitionString(
   }
 
   const states = definitionObj.States;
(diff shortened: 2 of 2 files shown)


src/span-link.ts
```diff
@@ -5,16 +5,12 @@ export function mergeStepFunctionAndLambdaTraces(
   resources: { [key: string]: GeneralResource },
   serverless: Serverless,
 ): void {
-  for (const resourceName in resources) {
-    if (resources.hasOwnProperty(resourceName)) {
```
avedmala, line 9: I'm also not sure when this would be false but I think we should check with @kimi-p, he might have some context

avedmala (commented):
Good refactor overall! I think we should just confirm that the safety check isn't needed

src/span-link.ts
kimi-p: I do not remember why I put I check there in the first place, but I think that should indeed be true. Thanks for simplifying the logic!

kimi-p (approved):

PR #528, opened 2024-09-23
feat: Support sfn->Lambda context injection when no Payload exists (case 1)

lym953:
### Current behavior:
In the Lambda step definition,
- If `Payload.$` doesn't exist, then inject context using `States.JsonMerge($$, $, false)`
- We don't check if `Payload` exists. The current behavior when `Payload` exists is that we also try to inject context using `States.JsonMerge($$, $, false)`. Then there would be an error when `serverless deploy` is run, saying `Payload` and `Payload.$` can't both exist, which can be confusing to users.

### What does this PR do?
Change context injection behavior so that if `Payload.$` or `Payload` don't exist, then inject context using:
```
$$['Execution', 'State', 'StateMachine']
```

This corresponds to Case 1 in the design doc [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit#bookmark=id.pczh1xpu3h9y)

#### Automated tests
Passed the existing and added tests

##### Steps
1. Create a State Machine with a Lambda step which has no Payload or Payload.$
```
InvokeLambdaStep:
  Type: Task
  Resource: arn:aws:states:::lambda:invoke
  Parameters:
    FunctionName:
      Fn::GetAtt: [yiming_plugin_hello, Arn]
    # Payload.$: $
  End: true
```

2. Run `serverless deploy`.
3. Check the State Machine definition in AWS UI

##### Result

Context was injected as expected:
```
    "InvokeLambdaStep": {
      "Type": "Task",
      "Resource": "arn:aws:states:::lambda:invoke",
      "Parameters": {
        "FunctionName": "arn:aws:lambda:sa-east-1:425362996713:function:yiming-serverless-demo-dev-yiming_plugin_hello",
        "Payload.$": "$$['Execution', 'State', 'StateMachine']"
      },
      "End": true
    }
```

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/step-functions-helper.spec.ts
+++ b/src/step-functions-helper.spec.ts
@@ -101,7 +101,7 @@ describe("test updateDefinitionString", () => {
     expect(newDefString).toContain("CONTEXT");
   });
 
-  it("test lambda step without Payload", async () => {
+  it("test lambda step without Payload or Payload.$", async () => {
     const definitionString = {
       "Fn::Sub": [
         '{"Comment":"fake comment","StartAt":"InvokeLambda","States":{"InvokeLambda":{"Type":"Task","Parameters":{"FunctionName":"fake-function-name"},"Resource":"arn:aws:states:::lambda:invoke","End":true}}}',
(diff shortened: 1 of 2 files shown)


src/step-functions-helper.spec.ts
```diff
@@ -112,10 +112,23 @@ describe("test updateDefinitionString", () => {
 
     const definitionAfterUpdate: StateMachineDefinition = JSON.parse(definitionString["Fn::Sub"][0] as string);
     expect(definitionAfterUpdate.States?.InvokeLambda?.Parameters?.["Payload.$"]).toBe(
-      "States.JsonMerge($$, $, false)",
+      "$$['Execution', 'State', 'StateMachine']",
     );
   });
 
+  it("test lambda step with custom Payload", async () => {
```
lym953, line 119: This is case 2. We don't inject context in this case for now. Will cover it later.

avedmala (approved):
🚢

PR #533, opened 2024-09-26
chore: Refactor sfn->sfn context injection code by merging two functions

lym953:
### What does this PR do?
Right now `updateDefinitionForStepFunctionInvocationStep()` calls `isSafeToModifyStepFunctionInvoctation()`. This PR merges the latter into the former.

### Motivation
Depending on the existing state, we will have slightly different instrumentation strategies. For example, if `Input` field doesn't exist, we will need to create this field then inject fields there. If `Input` already exists, then inject fields there directly.

Merging these two functions will make it more clear what we will do in each case.

### Testing Guidelines
Passed the existing tests.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/step-functions-helper.spec.ts
+++ b/src/step-functions-helper.spec.ts
@@ -1,8 +1,9 @@
 import {
   isDefaultLambdaApiStep,
-  isSafeToModifyStepFunctionInvoctation,
   StateMachineDefinition,
   updateDefinitionString,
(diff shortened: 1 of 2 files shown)


avedmala (approved):
thanks! 🚢

Issue #545, opened 2024-10-24
Not able to create high_error_rate monitor, 400 bad request


mariohoyos92:
## Expected Behavior
When using the `high_error_rate` recommended monitor, I would expect this to successfully create a monitor.

## Actual Behavior
When using the `high_error_rate` recommended monitor, I get `400 Bad Request: This could be due to incorrect syntax or a missing required tag for high_error_rate`. 

I believe this is happening because when we fetch the recommended monitors from datadog, there isn't anything that's returned that would get keyed as `high_error_rate` so the `isRecommendedMonitor` function returns `false` and we never set a `query` in the request body to create the monitor. 

I also think this line https://github.com/DataDog/serverless-plugin-datadog/blob/2d1e5f66bd950f5d0f4d50f2759de46ee4873e62/src/monitor-api-requests.ts#L214 doesn't really do anything, because after inspecting the payload we get from datadog the ids are keyed with `serverless-` not `serverless_`

## Steps to Reproduce the Problem
  1. Attempt to create a fresh high_error_rate recommended monitor

## Specifications

  - Serverless Framework version:  3.38.0
  - Datadog Serverless Plugin version:  5.72.0
  - Lambda function runtime (Python 3.7, Node 10, etc.):  Node 20
  
## Stacktrace
Here is what I logged out when calling the `getRecommendedMonitors` api

```
recommendedMonitors: {
    'serverless-[enhanced_metrics]_lambda_function_cold_start_rate_is_high': {
      name: 'High Cold Start Rate on $functionName in $regionName for $awsAccount',
      threshold: 0.2,
      message: 'More than 20% of the function's invocations were cold starts in the selected time range. This monitor uses an enhanced metric - to receive enhanced metrics, instrument your Lambda function with the [Datadog Lambda Extension](https://docs.datadoghq.com/serverless/libraries_integrations/extension/). Datadog's [enhanced metrics](https://docs.datadoghq.com/serverless/enhanced_lambda_metrics) and [distributed tracing](https://docs.datadoghq.com/serverless/distributed_tracing) can help you understand the impact of cold starts on your applications today. {{#is_alert}} \n' +
        ' Resolution: \n' +
        ' * Cold starts occur when your serverless applications receive sudden increases in traffic, and can occur when the function was previously inactive or when it was receiving a relatively constant number of requests. \n' +
[...]

PR #548, opened 2024-10-29
Identify recommended monitors with serverless_id tag

lym953:
### Background

Right now we are using a hard-coded map to identify the recommended monitor given its serverless_id like `high_error_rate`:
```
const recommendedMonitorApiIdToId = {
  "serverless-lambda_function_invocations_are_failing": "high_error_rate",
  ...
}
```

This is a temporary remediation for issue: https://github.com/DataDog/serverless-plugin-datadog/issues/545 and incident 31902.

### What does this PR do?

Use the tag from the Recommended Monitor API response to identify the monitor.

The response is like:
```
{
  "data": [
    {
      ...
        "name": "High Cold Start Rate on $functionName in $regionName for $awsAccount",
        "tags": [
          "serverless_id:high_cold_start_rate",
          "created_by:dd_sls_app"
        ],
    }
  ]
}
```

See full API response here: https://app.datadoghq.com/api/v2/monitor/recommended?count=50&start=0&search=tag%3A%22product%3Aserverless%22%20AND%20tag%3A%22integration%3Aamazon-lambda%22

### Motivation

To retrieve and filter recommended monitors in an elegant way, so we won't need to update our own map when we add a new recommended monitor.

#### Automated Testing
Passed the added test and existing tests.

##### Steps:
1. Update a stack with all the 7 recommended monitors. The stack had monitors `high_error_rate`, `timeout`, `high_cold_start_rate`, `high_throttles` before the update.
```
datadog:
    ...
    monitors:
      - high_error_rate:
          tags: ["team:serverless"]
      - timeout:
          tags: ["team:serverless"]
      - out_of_memory:
          tags: ["team:serverless"]
      - high_iterator_age:
          tags: ["team:serverless"]
      - high_cold_start_rate:
          tags: ["team:serverless"]
      - high_throttles:
          tags: ["team:serverless"]
      - increased_cost:
          tags: ["team:serverless"]
```
2. Run `serverless deploy`

##### Result:

All the 7 monitors have been updated or created. They all appear in Datadog App.
image
image

### Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/monitor-api-requests.ts
+++ b/src/monitor-api-requests.ts
@@ -22,7 +22,7 @@ export interface TemplateVariable {
   defaults: string[];
 }
 
-interface RecommendedMonitorParams {
+export interface RecommendedMonitorParams {
   id: string;
   attributes: {
     query: string;
@@ -34,21 +34,9 @@ interface RecommendedMonitorParams {
     };
     name: string;
     template_variables?: TemplateVariable[];
+    tags: string[];
   };
 }
-
-/**
- * Maps recommended monitor API IDs to IDs, where "API IDs" mean the keys in the monitor API response JSON
- */
-const recommendedMonitorApiIdToId = {
-  "serverless-lambda_function_invocations_are_failing": "high_error_rate",
-  "serverless-lambda_function_is_timing_out": "timeout",
-  "serverless-[enhanced_metrics]_lambda_function_is_running_out_of_memory": "out_of_memory",
-  "serverless-lambda_function's_iterator_age_is_increasing": "high_iterator_age",
-  "serverless-[enhanced_metrics]_lambda_function_cold_start_rate_is_high": "high_cold_start_rate",
-  "serverless-lambda_function_invocations_are_throttling": "high_throttles",
-  "serverless-[enhanced_metrics]_lambda_function_cost_is_increasing": "increased_cost",
-};
 
 export async function createMonitor(
   site: string,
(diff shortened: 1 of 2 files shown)


src/monitor-api-requests.ts
```diff
@@ -223,18 +211,18 @@ export async function getRecommendedMonitors(
...
-  recommendedMonitorApiId: string,
-): recommendedMonitorApiId is keyof typeof recommendedMonitorApiIdToId {
-  return recommendedMonitorApiId in recommendedMonitorApiIdToId;
+export function parseRecommendedMonitorServerlessId(recommendedMonitorParams: RecommendedMonitorParams): string {
+  for (const tag of recommendedMonitorParams.attributes.tags) {
+    if (tag.startsWith("serverless_id:")) {
+      return tag.substring(tag.indexOf(":") + 1);
+    }
+  }
+  throw new Error(`Recommended monitor ${recommendedMonitorParams.id} does not have a serverless_id tag.`);
```
clifordshelton, line 227: We should not be throwing an error here. The way it should be handled is, in the list of recommended monitors that get returned, we should look for `serverless_id` tag. If it is present, then we should return that template as a monitor that can be configured from plugin. If not, we should skip that template. Here's the reason behind it.

1. **This is not a hard requirement**: From the plugin, we're looking for all serverless monitor templates and then trying to see if there is a template that matches an `id` that user has specified. We could potentially create a monitor template that is meant to be only used from the "Browse Monitor Templates" view. In that case, it would be a lambda serverless monitor template that does not have a `serverless_id` tag.
2. **This check will cause all plugin deploys to fail if we make a mistake in our repo**: If we add a new lambda serverless monitor template that doesn't have `serverless_id` tag in it, all plugin deploys irrespective of whether the customer is using that specific monitor or not will fail because while parsing this error will get thrown.

lym953: Good point. Let me change it.

lym953: Updated and retested!

clifordshelton (approved):
Looks good 👍

Issue #584, opened 2025-03-18
Starting a step function inside a step function fails when passing Input.$ parameter with propagateUpstreamTrace turned on


meirav-navina:
## Expected Behavior
Start a step function execution from another step function with `Input.$` parameter while `propagateUpstreamTrace` flag is set to true.

## Actual Behavior
When trying to deploy the step function an `Cannot have both the field 'Input.$' and the field 'Input' at the same time` exception is thrown.

## Steps to Reproduce the Problem

  1. Set `propagateUpstreamTrace` to true in `serverless.yml`
  ```yaml
  custom:
    datadog:
      site: 'datadoghq.com'
      apiKeySecretArn: ...
      env: ${opt:stage}
      forwarderArn: ....
      enableStepFunctionsTracing: true
      propagateUpstreamTrace: true
   ``` 
  2. Add a startExecution state
  ```yaml
      state_name_1:
        Type: Task
        Resource: arn:aws:states:::states:startExecution.sync:2
        Parameters:
          StateMachineArn: arn:aws:states:us-east-1:123456:stateMachine:other_step_function_name_${opt:stage}
          Input.$: "$"
        OutputPath: "$.Output"
        ResultPath: "$"
        Next: another_state_name
  ```

## Specifications

  - Serverless Framework version:  3.31.0
  - Datadog Serverless Plugin version: 5.67.0
  - Lambda function runtime (Python 3.7, Node 10, etc.): Node 18
  
## Stacktrace
  
  ```
  Resource handler returned message: "Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: Cannot have both the field 'Input.$' and the field 'Input' at the same time at /States/state_name_1/Parameters, SCHEMA_VALIDATION_FAILED: Cannot have both the field 'Input.$' and the field 'Input' at the same time at /States/state_name_2/Parameters' (Service: AWSStepFunctions; Status Code: 400; Error Code: InvalidDefinition; Request ID: uuid1; Proxy: null)" (RequestToken: uuid2, HandlerErrorCode: InvalidRequest)
  ```

From what I can see there is no logic related to `Input.$` [here](https://github.com/DataDog/serverless-plugin-datadog/blob/main/src/step-functions-helper.ts#L252).

lym953:
Probably the same problem with https://github.com/DataDog/serverless-plugin-datadog/issues/517, which should have been fixed in latest versions. Could you try v5.71.0+?

[...]

PR #585, opened 2025-03-19
fix: Properly handle step function trace merging when Input.$ is set

lym953:
### Motivation

In step function definition, f the user sets `"Input.$` field for a step function invocation step and sets `propagateUpstreamTrace` flag, the plugin will try to add `Input` field to merge traces, causing `serverless deploy` to fail with the error:
> Cannot have both the field 'Input.$' and the field 'Input' at the same time

See https://github.com/DataDog/serverless-plugin-datadog/issues/584

### What does this PR do?

When `"Input.$` is set, do not try to set up trace merging. Instead, print a warning to direct the user to docs for manual setup.

### Testing Guidelines
1. Automated testing: Passed unit tests
2. Manual testing: Successfully deployed a stack that sets `"Input.$` field, which failed before.

### Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/step-functions-helper.spec.ts
+++ b/src/step-functions-helper.spec.ts
@@ -350,6 +350,12 @@ describe("test updateDefinitionForStepFunctionInvocationStep", () => {
     expect(updateDefinitionForStepFunctionInvocationStep(stepName, step, serverless, stateMachineName)).toBeFalsy();
   });
 
+  it('Case 0.4: Parameters field has "Input.$" field', async () => {
+    const parameters = { FunctionName: "bla", "Input.$": "$" };
+    const step = { Parameters: parameters };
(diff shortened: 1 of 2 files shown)


avangelillo (approved):
Is this a problem for other places where we have similar logic, like datadog-ci or the cdk constructs library?

lym953:
- For `datadog-ci`: we are already doing the same thing as in the PR
- For CDK, it's not a problem. We expose this API, where we get `input` as a dict, and we are free to add `CONTEXT` field there.
https://github.com/DataDog/datadog-cdk-constructs/blob/68a658463f845c4019e58a3272dd42a8a0c18ac0/src/span-link.ts#L48

PR #614, opened 2025-09-30
feat(appsec): deprecate enableASM in favor of `appSecMode`

florentinl:
### What does this PR do?

This PR deprecates the `enableASM` configuration option in favor of the new `appSecMode` option.

The Datadog Lambda Library for Python (version `8.114.0` or later) can run App and API Protection directly inside the instrumented application, giving the security engine additional context. Support for other runtimes will follow. This complements the in-extension implementation delivered by the Datadog Lambda Extension. `appSecMode` selects which implementation runs:
- `on`: with `addLayers: true` on Python functions, App and API Protection runs within the library; other runtimes or manually supplied libraries fall back to the extension.
- `tracer`: always use the library implementation(every function must be in Python and use the Python library at version `8.114.0` or newer) => this is a "force" option in case users install the library themselves or use a Docker Image etc... we can't check that the conditions are met.
- `extension`: always use the extension implementation, even when the library is present.
- `off`: disable App and API Protection entirely.

### Motivations

- ASM (Application Security Management) is no longer the name of the product as it has been renamed to AAP (Application and API Protection), keeping `enableASM` does not make much sense anymore. The `appSecMode` semantic resembles the env var and is more explicit.

- We want to push re-using the python tracer in lambda to deliver the AAP features instead of within the runtime api proxy of the extension. This requires however using `datadog-lambda-python` with a version that satisfies `>=8.114.0`.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -275,6 +275,12 @@ export function setEnvConfiguration(config: Configuration, handlers: FunctionInf
         );
       }
 
+      if (config.appSecMode === "tracer" && !IN_TRACER_APPSEC_SUPPORTED_RUNTIMES.includes(type)) {
+        throw new Error(
+          `\`appSecMode\` is set to \`tracer\`. This requires the function to be a Python runtime. Found: ${type}`,
+        );
+      }
+
       if (
         (config.appSecMode === "on" && IN_TRACER_APPSEC_SUPPORTED_RUNTIMES.includes(type) && config.addLayers) ||
         config.appSecMode === "tracer"
--- a/README.md
+++ b/README.md
@@ -209,11 +209,17 @@ custom:
 
 ### Enable in-tracer App and API Protection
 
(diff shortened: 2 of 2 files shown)


avara1986 (approved):

src/env.ts
```diff
@@ -250,12 +255,44 @@ export function setEnvConfiguration(config: Configuration, handlers: FunctionInf
...
+        throw new Error("`enableAppSec` requires `enableDDTracing` to be enabled");
+      }
+      if (type === RuntimeType.PYTHON) {
+        environment[ddAppsecEnabledEnvVar] ??= config.enableAppSec;
+      } else {
+        environment[AWS_LAMBDA_EXEC_WRAPPER_VAR] ??= AWS_LAMBDA_EXEC_WRAPPER;
+        environment[ddServerlessAppsecEnabled] ??= config.enableAppSec;
+      }
+    }
+    if (config.enableAppSecRuntimeApiProxy !== undefined && config.enableAppSecRuntimeApiProxy) {
```
TalUsvyatsky, line 280: Would this work too?
```suggestion
    if (config.enableAppSecRuntimeApiProxy) {
```

florentinl: Absolutely.

PR #695, opened 2026-04-14
chore: remove monitors test from pr workflow

ava-silver:
### What does this PR do?

1. Removes the \"Test getting recommended monitors\" step from the PR integration-tests workflow.
2. Renames `integration_tests/` → `snapshot_tests/` (directory, script, workflow, and config references).

### Motivation

**Monitors test removal:** This step hits `api.datadoghq.com/api/v2/monitor/recommended` and asserts that certain named recommended monitors exist. It tests nothing in this repo -- those monitors are configured server-side by Datadog's backend team, so no PR here can affect whether this test passes or fails.

The test was causing every Dependabot PR to fail on its first CI run because Dependabot PRs don't have access to regular repo secrets (`DD_API_KEY`/`DD_APP_KEY`), only the restricted Dependabot secret scope. The fix in #692 addressed a separate CloudFormation credentials issue but missed this one.

The test already has a dedicated home in `monitor_api_integration_test.yml`, which runs on a daily cron with a Slack alert on failure -- the right place for a canary that monitors external API availability.

**Rename:** The tests in `integration_tests/` run `sls package`, generate a CloudFormation template, and diff it against checked-in snapshots. They never deploy anything and require no real AWS connectivity. Calling them "integration tests" implies live AWS interaction is expected -- which is exactly why the credential failure looked like a bug rather than an obvious non-issue. Renaming to `snapshot_tests` makes the nature of the tests self-evident.

Note: `integrationTesting` / `testingMode` config flags in `src/env.ts`, `src/forwarder.ts`, `src/index.ts` are a different concept (bypass forwarder ARN validation) and are left untouched.

### Testing Guidelines

CI should pass on Dependabot PRs without any manual intervention.

### Additional Notes

Fixes the remaining flakiness on Dependabot PRs not addressed by #692.

### Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

### Check all that apply


ava-silver:
* **#695** <a href="https://app.graphite.com/github/pr/DataDog/serverless-plugin-datadog/695?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/serverless-plugin-datadog/695?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

nina9753 (approved):

.github/workflows/snapshot_tests.yml
```diff
@@ -35,10 +35,5 @@ jobs:
...
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-        run: ./scripts/run_integration_tests.sh
+        run: ./scripts/run_snapshot_tests.sh
 
-      - name: Test getting recommended monitors
-        env:
-          DD_API_KEY: ${{ secrets.DD_API_KEY }}
-          DD_APP_KEY: ${{ secrets.DD_APP_KEY }}
-        run: npm run test:integration
```
avangelillo, line 44: I'll be honest, I never really looked that closely into what these tests were doing but I assumed that it was deploying to AWS then pulling the deployed template down and asserting that it is both deployable and matches the snapshot.  [In my opinion] just doing a snapshot test like this isn't really all that useful, and trying to add a deploy step here is worth doing if we have the credentials available.

> It tests nothing in this repo -- those monitors are configured server-side by Datadog's backend team, so no PR here can affect whether this test passes or fails.
> ...
> The test already has a dedicated home in monitor_api_integration_test.yml, which runs on a daily cron with a Slack alert on failure -- the right place for a canary that monitors external API availability.

I think I've added a monitor to this list before, but if we don't really own it I'd go even further and say that we shouldn't be the ones trying to test it.

Since the plugin does get the recommended monitors from an API, then [creates / updates / deletes them](https://github.com/DataDog/serverless-plugin-datadog/blob/main/src/monitors.ts#L194) based on the response, I think the including that in (actual) integration tests that run as a part of a PR is still the appropriate place to run these tests, if a PR breaks that we should not merge it in and we're missing that validation.  Maybe not by specific names like how the test works now, but that something was created.

We also need a Datadog API / app key for the remote instrumentation tests, and they are [stored in Secrets Manager](https://d-906757b57c.awsapps.com/start/#/console?account_id=425362996713&role_name=account-admin-8h&redirect=true&destination=https%3A%2F%2Feu-north-1.console.aws.amazon.com%2Fsecretsmanager%2Flistsecrets%3Fregion%3Deu-north-1) in the sandbox account and [retrieved during the integration tests](https://github.com/DataDog/serverless-remote-instrumentation/blob/prod/integration-tests/utilities/datadog-keys.ts), I would recommend doing something similar to that if there is a problem storing the keys in GitHub.

avangelillo (approved):

PR #715, opened 2026-07-14
feat: support osls 4

stefan-person:
### What does this PR do?

Support osls 4 and use AWS SDK v3 for osls 4 but stick to v2 for the existing serverless or osls 3 case

### Motivation

We want to upgrade to osls 4 soon and the serverless-plugin-datadog is currently not supporting it. So I thought to give it a try

### Testing Guidelines

Created a small test-lambda and deployed it once with osls 4 and once with osls 3. For both I get the log entries on datadog side for START, END and REPORT.

### Additional Notes

To run the tests on my windows machine, I had to adjust the jest config a bit. It should now work for win and unix environment

### Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### Check all that apply

- [X] This PR's description is comprehensive
- [X] This PR's changes are covered by the automated tests

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,6 +9,6 @@ module.exports = {
   testEnvironment: "<rootDir>/testEnvironment.js",
   clearMocks: true,
   collectCoverageFrom: ["src/**/*.ts"],
-  testRegex: "(src\\/).*(\\.spec\\.ts)$",
-  testPathIgnorePatterns: ["\\.snap$", "<rootDir>/node_modules/", "snapshot_tests/"],
+  testRegex: String.raw`(src[\/]).*(\.spec\.ts)$`,
+  testPathIgnorePatterns: [String.raw`\.snap$`, "<rootDir>/node_modules/", "snapshot_tests/"],
 };
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,6 +9,6 @@ module.exports = {
   testEnvironment: "<rootDir>/testEnvironment.js",
   clearMocks: true,
   collectCoverageFrom: ["src/**/*.ts"],
-  testRegex: String.raw`(src[\/]).*(\.spec\.ts)$`,
+  testRegex: String.raw`(src[\\/]).*(\.spec\.ts)$`,
   testPathIgnorePatterns: [String.raw`\.snap$`, "<rootDir>/node_modules/", "snapshot_tests/"],
 };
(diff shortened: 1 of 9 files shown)


src/aws-sdk-request.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+  await aws.request("Lambda", "getFunction", { FunctionName: functionName });
+}
+
+/** Calls CloudWatchLogs DescribeSubscriptionFilters for a log group. */
+export async function cloudWatchLogsDescribeSubscriptionFilters(aws: Aws, logGroupName: string): Promise<any[]> {
+  if (supportsAwsSdkV3(aws)) {
+    const { CloudWatchLogsClient, DescribeSubscriptionFiltersCommand } =
+      await import("@aws-sdk/client-cloudwatch-logs");
+    const client = new CloudWatchLogsClient(await aws.getAwsSdkV3Config());
```
ava-silver: Could we cache and reuse these SDK clients, ideally by provider, service, and region? This helper is called once per eligible log group, so a deployment with many log groups will create many clients, config resolutions, and connection pools. osls's SDK v3 guidance also recommends caching client promises.

ava-silver (commented):
Thanks for the PR, and sorry it took us so long to review this! The core compatibility approach looks sound, and the build, tests, lint, and formatting all pass locally. I left two non-blocking notes around AWS client reuse and the Windows Jest regex change. If you have bandwidth, go ahead and implement them, or I can handle them before merging, once I finish an end-to-end test of the changes!

jest.config.js
```diff
@@ -9,6 +9,6 @@ module.exports = {
   testEnvironment: "<rootDir>/testEnvironment.js",
   clearMocks: true,
   collectCoverageFrom: ["src/**/*.ts"],
-  testRegex: "(src\\/).*(\\.spec\\.ts)$",
-  testPathIgnorePatterns: ["\\.snap$", "<rootDir>/node_modules/", "snapshot_tests/"],
+  testRegex: String.raw`(src[\/]).*(\.spec\.ts)$`,
```
ava-silver: The PR description says this is intended to support Windows paths, but `[\/]` as written here still matches only `/`; `String.raw` just preserves that pattern. To match both separators, this needs `[\\/]`. If this is only a readability change, the behavior is fine, but the stated Windows fix is not currently achieved.

ava-silver (approved):
Thanks again for your help here!

